PR #1, opened 2024-05-21
feat: Color.toArray()

amyspark-ng:
Added color.toArray() since thought i'd be a nice feature<br>
`ex: obj.color.toArray -> [255, 255, 255]`
Re-did the PR on the COOLER kaboom 😎


niceEli:
LGTM!

Issue #6, opened 2024-05-22
feat: onClick 'mouseButton' parameter to detect what mouseButton was pressed


lajbel:

mflerackers:
I think using [isMousePressed](https://kaboomjs.com/#isMousePressed)(button?: MouseButton) you can check whether the click was triggered using a certain button. But this may be more cumbersome.

lajbel:
I think it changes the style of the user, passing from `onClick()` to `onUpdate(() => { if(isMousePressed) }), so we should have parity between event and if functions

lajbel:
If version:

```js
onUpdate(() => {
    if(isMousePressed("right")) {
        console.log("right");
    }
});
```

event version:
```js
onClick((btn) => {
    console.log(btn); // right
});
```

lajbel:
fixed on: 5e16da1

PR #10, opened 2024-05-22
feat: add button choice to onClick

niceEli:
adds #6


niceEli:
after looking at it multiple times i think its good. its not that complex, only 3 lines changed

PR #20, opened 2024-05-22
refactor: separate components in files

lajbel:
This is a test PR, to see how it goes, but we have to have a way to have a "global" KaboomCtx, and it should be used in all files so we can separate components and call these functions.

Another approach will be simply separate all functions and only use the state "app"

- [ ] Find a way to separate kaboom methods that depends on another kaboom methods easily


niceEli:
Is this test or should merge?

lajbel:
Shouldn't

amyspark-ng:
Maybe should move components in categories because having every component in a different file feels bloating

mflerackers:
transform: pos, rotate, scale
draw: sprite, rect, circle, polygon
physics: area, body
animation: timer

amyspark-ng:
yes, yes

SpcFORK:
> Find a way to separate kaboom methods that depends on another kaboom methods easily

Maybe switch to classes?

PR #21, opened 2024-05-22
chore: delete docs folder

amyspark-ng:
deleted docs folder


niceEli:
I think this is docs type not chore but I could be wrong

lajbel:
it should be refactor (jk)

nah chore is good, docs too

PR #36, opened 2024-05-23
refactor: wait and loop

### LINKED ISSUE #35: refactor: rewrite loop to use update directly
make it less cursed.
add a choice to wait for execute or just run right away.
might be breaking change but may not depending on implementation

niceEli:
fixes #35

ran dprint fmt so it formatted my changes


niceEli:
will fix

PR #38, opened 2024-05-23
fix: area displays size on inspect

amyspark-ng:
area apparently used to display "area" when inspecting, now it displays the vector size, in the same format as pos

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -3847,6 +3847,10 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
                     return area.transform(game.cam.transform);
                 }
             },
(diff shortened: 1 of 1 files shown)


mflerackers:
Most likely you'll need to add inspect methods to each shape class. Then you can add `shape: ${shape.inspect()}`

amyspark-ng:
that'd be nice, i could add that too

amyspark-ng:
think i won't since it looks like the debug.inspect system is getting remade into something much nicer, feels like adding shape might imply also adding things like offset maybe? and also bloat the little box you get when you hover something on inspect mode, so i don't think i will

PR #56, opened 2024-05-26
feat: added getCurrentScene function

amyspark-ng:
Added a function that gets the name of the current scene, tried to avoid adding a property to the game object but i really couldn't find a way to do it, so i just had to add one

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1325,8 +1325,8 @@ export interface KaboomCtx {
     /**
      * Gets the name of the current scene.
      *
-     * @since v3000.0
-     * @group Scenes
+     * @since v3001.0
+     * @group Scene
      */
     getSceneName(): string;
     /**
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -1321,6 +1321,14 @@ export interface KaboomCtx {
      * @group Events
      */
     onSceneLeave(action: (newScene?: string) => void): EventController;
+    /**
+    /**
+     * Gets the name of the current scene.
+     *
+     * @since v3000.0
```
lajbel: 3001 in fact

src/types.ts
```diff
@@ -1321,6 +1321,14 @@ export interface KaboomCtx {
...
+     * @group Scenes
```
lajbel: Also current used is Scene

lajbel (changes requested):

lajbel (approved):

niceEli (approved):
lgtm

PR #59, opened 2024-05-26
feat: added debugKey

amyspark-ng:
Added a property for kaboomOpt where you can set the key that toggles the debug mode, couldn't find where the defaults in kaboomOpt were defined so i just checked in the line that does the key checking
`app.onKeyPress(!gopt.debugKey ? "f1" : gopt.debugKey, () => debug.inspect = !debug.inspect);`

closes: https://github.com/marklovers/kaplay/issues/46

### DIFF
--- a/src/kaboom.ts
+++ b/src/kaboom.ts
@@ -4883,7 +4883,7 @@ export default (gopt: KaboomOpt = {}): KaboomCtx => {
         });
 
         if (gopt.debug !== false) {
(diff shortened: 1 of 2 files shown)


src/kaboom.ts
lajbel, line 1: you can simplify with `gopt.debugKey ?? "f1"`

amyspark-ng: nice to learn more syntax

PR #62, opened 2024-05-27
refactor: added fadeIn to opacityComp

amyspark-ng:
I changed fadeIn from being a whole component to be a function of the opacity comp, thought it was more intuitive and made more sense.

Might be a bit of a breaking change


niceEli:
this is breaking change

PR #65, opened 2024-05-28
feat: Bevel, round and miter line join styles

mflerackers:
Implements bevel, round and miter line join styles
TODO: miter limit


lajbel:
should I merge? Or it's a draft PR

PR #68, opened 2024-05-28
refactor: rename `kaboom`/`startGame` to `kaplay`

lajbel:
Added `kaplay()`

the kaboom example teach how you can still use `kaboom`


niceEli (approved):
LGTM

PR #73, opened 2024-05-29
chore: make tests use js modules

niceEli:
also fixed raycastShape or else the tests would fail (thats how strict the tests work now)


niceEli:
wait to merge. there seems to be a issue only on actions (it works on my machine)

niceEli:
somehow still broken

PR #89, opened 2024-06-03
feat: SpriteComp.getCurAnim()

### LINKED ISSUE #88: feat: `SpriteComp.getCurAnim()` for get the current animation data
As title says, there should be a method to get all animation data, for example you can use `SpriteComp.curAnim()` to get the current animation name, but you can't get if the animation is `pingpong`, `speed` or other props

lajbel:
close #88


niceEli (approved):
LGTM

PR #90, opened 2024-06-03
feat: inspecting outline displays properties

amyspark-ng:
now when debug inspecting, the outline component will show width of outline and color of outline

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - added `getSceneName()` to get the current scene name
 - added support for gravity direction
 - added `SpriteComp.getCurAnim()` to get the current animation name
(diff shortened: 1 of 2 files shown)


lajbel:
Can you show an image?

amyspark-ng:
## before:
image

## after: 
image

lajbel:
Awesome, add changelog

amyspark-ng:
is it really necessary, some comps had their properties shown in the inspect window recently such as area showing scale or shader showing shader name but they haven't been added to changelog

niceEli (approved):
👍🏻

amyspark-ng:
oh nice, thank you eli maybe we should add the changelog for all the other comps too

PR #108, opened 2024-06-04
feat: Add using Function return values.

SpcFORK:


niceEli:
nvm i spent more time looking at it. it seems good

niceEli (approved):

PR #117, opened 2024-06-07
fix: full screen judgment

luluxia:
The `isFullscreen()` function gets the fullscreen DOM object, but doesn't determine if it's the game canvas or not. When the `letterbox` option is set in the game, going fullscreen on the entire web page will result in the game canvas size changing, but the game resolution not updating, so I've changed the `isFullscreen()` function to work around this issue.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -189,12 +189,6 @@ export default (opt: {
         else if (document.webkitExitFullScreen) document.webkitExitFullScreen();
     }
 
-    function getFullscreenElement(): Element | void {
-        return document.fullscreenElement
-            // @ts-ignore
-            || document.webkitFullscreenElement;
-    }
-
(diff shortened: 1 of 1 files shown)


src/app.ts
```diff
@@ -204,7 +198,9 @@ export default (opt: {
     }
 
     function isFullscreen(): boolean {
-        return Boolean(getFullscreenElement());
+        return document.fullscreenElement === state.canvas
+            // @ts-ignore
+            || document.webkitFullscreenElement === state.canvas;
```
niceEli, line 203: does this only work with safari and safari like browsers (WebKit)

niceEli: oh wait nvm i see, i was confused from indentation and that its a || operator

niceEli (approved):

PR #130, opened 2024-06-12
feat: added tags array

amyspark-ng:
code wrote by MF, so cool


niceEli:
would this need to be in types.ts aswell?

niceEli (approved):
👍🏻

PR #137, opened 2024-06-13
feat: named component to assign a name, and query, a more versatile way to get certain objects

mflerackers:


src/types.ts
```diff
@@ -3581,6 +3589,12 @@ export interface GameObjRaw {
      * @since v3000.0
      */
     get(tag: Tag | Tag[], opts?: GetOpt): GameObj[];
+    /**
+     * Get a list of all game objs with certain properties.
+     *
+     * @since v3001.0
+     */
+    query(opt: QueryOpt): GameObj[];
```
lajbel, line 3597: Looks good, but it's query the best option? What about `getQuery` (because is a like get and maybe it replaces get in the future) or `getWithQuery`

src/types.ts
```diff
@@ -677,6 +678,13 @@ export interface KaboomCtx {
      * @group Components
      */
     lifespan(time: number, options?: LifespanCompOpt): EmptyComp;
+    /**
+     * Names an object.
+     *
+     * @since v3001.0
+     * @group Components
+     */
+    named(name: string): NamedComp;
```
lajbel, line 687: `name()` sounds better for me

src/components/misc/named.ts
```diff
@@ -0,0 +1,7 @@
+import type { Comp, NamedComp } from "@/types";
```
lajbel, line 1: Also if you want you can put the types here, then do 

```js
// types.ts
import type { NamedComp } from "@/components"
export { NamedComp }
```

src/kaboom.ts
```diff
@@ -3598,6 +3600,95 @@ const kaplay = (gopt: KaboomOpt = {}): KaboomCtx => {
...
+                            function recurse(child: GameObj) {
+                                return [
+                                    child,
+                                    ...child.children.flatMap(recurse),
+                                ];
+                            },
+                        );
+                        break;
+                }
+                console.log(`hierarchy: ${hierarchy}`, list);
```
lajbel: debug console log here

src/kaboom.ts
```diff
@@ -3598,6 +3600,95 @@ const kaplay = (gopt: KaboomOpt = {}): KaboomCtx => {
...
+                        // Accept if all match
+                        list = list.filter(o => o.is(opt.include));
+                    } else { // includeOp == "or"
+                        // Accept if some match
+                        list = list.filter(o =>
+                            (opt.include as string[]).some(t => o.is(t))
+                        );
+                    }
+                }
+                console.log("include", list);
```
lajbel: debug console log here

src/kaboom.ts
```diff
@@ -3598,6 +3600,95 @@ const kaplay = (gopt: KaboomOpt = {}): KaboomCtx => {
...
+                        );
+                    } else { // distanceOp === "far"
+                        list = list.filter(o =>
+                            o.pos && this.pos.sdist(o.pos) > sdist
+                        );
+                    }
+                }
+                console.log(
+                    `distance: ${opt.distance}, distanceOp: ${opt.distanceOp}`,
+                    list,
```
lajbel: debug console log here

src/types.ts
mflerackers: I think query is good. It's short and tells you what it does, it queries the scene for objects. I first thought about just adding a GetOpt as third kind of parameter to get, but it would not go well with the options already passed to get (like recursive).
getQuery sounds like getting a query. The first verb should say what happens, so I think query is best.

src/components/misc/named.ts
mflerackers: I actually wanted name instead of named. But it wouldn't let me. I did the NamedComp originally inside the component file, however I wanted to see an example of the new convention first, so it would be the same.

src/types.ts
lajbel: good

lajbel (approved):

PR #150, opened 2024-06-17
refactor: move component types/opt from types.ts to their modules

lajbel:
- moves components interfaces/opt to their modules in `components/`
- removes getInternalContext in favour of `const {} = k._k`
- moved raycast to components (again)
- this doesn't remove even the 10% of the line count in types.ts


mflerackers (approved):
Looks good 👍

PR #154, opened 2024-06-18
feat: camFlash()

amyspark-ng:
Adds camFlash, takes as a parameter color (WHITE) and a fade out time (1 second), closes https://github.com/marklovers/kaplay/issues/153


mflerackers (approved):
Looks good

PR #162, opened 2024-06-19
chore: re-added missing examples

amyspark-ng:
not sure if i had to add ALL of them or only the ones that were before


niceEli (approved):
👍🏻, i will say this is going to make test passing take longer tho.

amyspark-ng:
well it's fair since they were bound to be added and are good

PR #170, opened 2024-06-24
feat: support for custom names in inspect 

amyspark-ng:
would fix https://github.com/marklovers/kaplay/issues/168, i think i'd be better behaviour

sorry for closing and reopening and closing and reopening im struggling with git right now

### DIFF
--- a/examples/inspectExample.js
+++ b/examples/inspectExample.js
@@ -1,0 +1,31 @@
+kaplay()
+
+// # will delete this file when changes get merged/declined i don't intend this to be an actual example
+function customComponent() {
+	return {
+		id: "compy",
+		customing: true,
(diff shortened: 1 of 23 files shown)


amyspark-ng:
i made some changes like

- now when both axis (x and y) of an area/scale component are the same it shows up as `area: 1.0x`, when they're different it shows up as `area: (1.0x, 1.2y)`
- now body returns on inspect gravityScale as `gravityScale: 1x`
- pos returns on inspect like `pos: (100x, 300y)`
- rect returns on inspect like: `rect: (100w, 300h)`

if any of these changes are not approved notify me and i'll revert

niceEli (approved):
very cool update, allows for better debugging, also thats a ton of files changed, doing a S.A.M rn

PR #175, opened 2024-06-25
feat: comps navigation mesh, sentry and patrol components, particle system, random from  shape

mflerackers:
Feature still in progress, this is not the final API


src/components/draw/particles.ts
```diff
@@ -0,0 +1,197 @@
...
+        this.lt = lt;
+        this.gc = false;
+    }
+
+    get progress() {
+        return this.t / this.lt;
+    }
+}
+
+export type ParticlesOpt = {
```
lajbel, line 35: this opt looks kinda redundant, I mean, speed, not particleSpeed

lajbel (approved):
I think looks fine for me, only that details

niceEli:
theres conflicts caused by #176

lajbel (approved):

PR #176, opened 2024-06-25
chore: rm dup types, descriptions, categories

lajbel:
this pr remove some duplications between gfx, math and types.ts, also moves some cateogires


niceEli (approved):
lgtm

PR #177, opened 2024-06-25
chore: simplify pr template

lajbel:
## Description

I think the current PR template is kinda annoying, I don't even like to use this, I changed to this, I would like to check if other people likes it


amyspark-ng:
i personally think is fun but it's not practical

lajbel:
For example, the part of what kind of PR is, is not necessary, that's why we use conventional commit names

niceEli (approved):
👍🏻

amyspark-ng:
makes me a little sad because eli made it

PR #180, opened 2024-06-27
chore: remove matter package

lajbel:
## Description

Because is not supported and not updated. Maybe will back on other time, in other repo

### Issues or related

- Related Issue #
- Closes #


niceEli (approved):
its good

PR #191, opened 2024-07-03
feat: text inputs

niceEli:
## Description

this allows you to make text input boxes, with this feature you can make any type of ui you want: search boxes, sign up panels, and making a discord client.

### Issues or related

- Related Issue #N/A
- Closes #N/A

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -42,6 +42,7 @@ import type {
     StayComp,
     TextComp,
     TextCompOpt,
+    TextInputComp,
     TileComp,
     TileCompOpt,
     TimerComp,
@@ -135,6 +136,7 @@ export type {
     StayComp,
     TextComp,
     TextCompOpt,
+    TextInputComp,
     Texture,
     TileComp,
     TileCompOpt,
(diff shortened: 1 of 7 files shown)


src/types.ts
```diff
@@ -699,6 +701,23 @@ export interface KaboomCtx<
      * @group Components
      */
     shader(id: string, uniform?: Uniform | (() => Uniform)): ShaderComp;
+    /**
+     * Enable timer related functions like wait(), loop(), tween() on the game object.
```
mflerackers: I guess this needs to be Enables text input on an object with a text component.

src/types.ts
```diff
@@ -699,6 +701,23 @@ export interface KaboomCtx<
...
+     *
+     * @example
+     * ```js
+     * const obj = add([
+     *     text(""),
+     *     textInput(),
+     * ])
+     *
+     * obj.hasFocus = false
+     * debug.log(obj.text) // oh no i cant see my new text since it was disabled
```
mflerackers, line 715: You just can't input text right? You can still see it.

mflerackers (approved):

src/types.ts
niceEli: yes, when hasFocus is false then you cant put new text, you can still see all the old text

PR #196, opened 2024-07-06
fix: unuse allows for removal of needed components

### LINKED ISSUE #126: feat: check type in return of `resolveShader`
```js
    function resolveShader(
        src: RenderProps["shader"],
    ): ShaderData | Asset<ShaderData> | null {
        if (!src) {
            return gfx.defShader;
        }
        if (typeof src === "string") {
            const shader = getShader(src);
            if (shader) {
                return shader.data ?? shader;
            } else if (loadProgress() < 1) {
                return null;
            } else {
                throw new Error(`Shader not found: ${src}`);
            }
        } else if (src instanceof Asset) {
            return src.data ? src.data : src;
        }
        // TODO: check type
        // @ts-ignore
        return src;
    }
```

### LINKED ISSUE #165: bug: unuse allows for removal of needed components
Unuse does not check dependencies before removing a component.

lajbel:
## Description
It check all components in `unuse()` searching for dependents. Also I added dprint vscode extension to devcontainer, something necessary.

### Issues or related
- Closes #165

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,28 +1,24 @@
-
 // For format details, see https://aka.ms/devcontainer.json. For config options, see the
 // README at: https://github.com/devcontainers/templates/tree/main/src/javascript-node
 {
 	"name": "KAPLAY",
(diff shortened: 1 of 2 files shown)


niceEli (approved):
👍

PR #199, opened 2024-07-06
fix: Checks implicitly ensure values 

SpcFORK:
## Description

Use `!` for implicitly not falsey on canvas size check.

### Issues or related

Causes type issues.


lajbel:
why? in that case there's always a width and height

lajbel (approved):

PR #207, opened 2024-07-06
fix: If shader error is malformed, throw

SpcFORK:
## Description

If shader error is malformed, throw

### Issues or related

Ensure types.
Type collision.


lajbel:
format

lajbel (approved):

PR #213, opened 2024-07-07
fix: Ensure required properties exist for `drawUVQuad`

SpcFORK:
## Description

Ensure required properties exist for `drawUVQuad`.

### Issues or related

Type collision.


SpcFORK:
image
image

---
It is likely that using `null` somewhere in the codebase is a bug, and should be handled with `undefined` instead rather than having two falsey types where one is technically defined.

Maybe somebody wrote a null somewhere, is probably a good idea to refactor if possible.

SpcFORK:
Tests have failed because it seems `drawUVQuad` doesn't need these, so then the push functions should accept undefined.
image

mflerackers (changes requested):
Pos, angle and scale aren't needed, this will break a lot of code. pushTranslate, pushRotate and pushScale are NOP if the params are null.

SpcFORK:
They are gone now, and have been switched for `!`

PR #216, opened 2024-07-07
fix: `opt` properties are possibly undefined.

SpcFORK:
## Description

opt properties are possibly undefined, using `!` fixes this.


mflerackers (changes requested):
But why check pos when pushTranslate checks for undefined?

SpcFORK:
?

No checks are being made.

mflerackers:
Yes, ! basically says trust me bro, this isn't going to be null. Which is not the correct solution. Your problem is that even though pushTranslate checks for undefined, it's signature is pushTranslate(...args: Vec2Args), and Vec2Args is

export type Vec2Args =
    | [number, number]
    | [number]
    | [Vec2]
    | [number | Vec2]
    | [];

so null is not included. The correct solution is to fix the signature of pushTranslate.

PR #246, opened 2024-07-09
refactor: split utils in various files

lajbel:
## Description

Simply split all utils.ts content in a utils folder.


niceEli (approved):
how was this only 2 commits

lajbel:
easy than other refactors

PR #251, opened 2024-07-10
feat: allow for getting center of basic display objects

SpcFORK:
## Description

feat: allow for getting center of basic display objects

### DIFF
--- a/src/components/draw/circle.ts
+++ b/src/components/draw/circle.ts
@@ -1,6 +1,6 @@
 import { getKaboomContext } from "../../kaboom";
 import { Rect, Vec2 } from "../../math";
-import type { Comp, GameObj } from "../../types";
+import type { Comp, GameObj, KaboomCtx } from "../../types";
 import type { AnchorComp } from "../transform/anchor";
 import type { outline } from "./outline";
 
--- a/src/components/draw/circle.ts
+++ b/src/components/draw/circle.ts
@@ -38,7 +38,11 @@ export interface CircleCompOpt {
     fill?: boolean;
 }
 
-export function circle(radius: number, opt: CircleCompOpt = {}): CircleComp {
+export function circle(
+    this: KaboomCtx,
+    radius: number,
+    opt: CircleCompOpt = {},
+): CircleComp {
     const k = getKaboomContext(this);
     const { getRenderProps } = k._k;
 
--- a/src/components/draw/circle.ts
+++ b/src/components/draw/circle.ts
@@ -1,6 +1,6 @@
 import { getKaboomContext } from "../../kaboom";
 import { Rect, Vec2 } from "../../math";
-import type { Comp, GameObj, KaboomCtx } from "../../types";
+import type { Comp, GameObj } from "../../types";
 import type { AnchorComp } from "../transform/anchor";
 import type { outline } from "./outline";
 
@@ -19,6 +19,10 @@ export interface CircleComp extends Comp {
      * @since v3000.0
      */
     renderArea(): Rect;
+    /**
+     * Center of circle
+     */
+    center(): Vec2;
 }
 
 /**
@@ -34,11 +38,7 @@ export interface CircleCompOpt {
     fill?: boolean;
 }
 
-export function circle(
-    this: KaboomCtx,
-    radius: number,
-    opt: CircleCompOpt = {},
-): CircleComp {
+export function circle(radius: number, opt: CircleCompOpt = {}): CircleComp {
     const k = getKaboomContext(this);
     const { getRenderProps } = k._k;
 
@@ -57,6 +57,9 @@ export function circle(
                 this.radius * 2,
                 this.radius * 2,
             );
+        },
+        center() {
+            return new Vec2(this.radius, this.radius)
         },
         inspect() {
             return `radius: ${Math.ceil(this.radius)}`;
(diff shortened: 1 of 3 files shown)


mflerackers:
These don't take into account the offset of the anchor. Also center of circle would be 0,0.

SpcFORK:
> These don't take into account the offset of the anchor. Also centre of circle would be 0,0.

I don't believe having the anchor will change getting the relative centre of an object.
Unless there is some benefit to this, the current way I have implemented allows for quickly using it for relative vector adding and stuff.

mflerackers:
If the anchor is top-left, the center of the rect is is width/2, height/2, but if the anchor is center, the center is 0, 0.
While a rect is drawn starting at pos, extending width, height down-right, a circle is drawn with it's center at pos, extending top-left and bottom right by radius pixels.
So what exactly is the center you are calculating?

mflerackers:
What you probably want is 

```ts
const rect: Rect = this.renderArea();
return rect.center();
```

SpcFORK:
> If the anchor is top-left, the center of the rect is is width/2, height/2, but if the anchor is center, the center is 0, 0. While a rect is drawn starting at pos, extending width, height down-right, a circle is drawn with it's center at pos, extending top-left and bottom right by radius pixels. So what exactly is the center you are calculating?

Center of the shape? Pretty sure it wouldn't need a pivot calculation to get the absolute center of the drawn shape based off of half of the parameter width given, not the pivot center of the shape

mflerackers:
This won't work for text, polygon or uvquad.
I think the correct solution for your problem is
```ts
function getCenterRelativeToTopLeft(object) {
    if (object.renderArea) {
        const bbox = object.renderArea().bbox();
        return bbox.center().sub(bbox.pos);
    }
    else {
        return null
    }
}
```

SpcFORK:
> This won't work for text, polygon or uvquad. I think the correct solution for your problem is
> 
> ```ts
> function getCenterRelativeToTopLeft(object) {
>     if (object.renderArea) {
>         const bbox = object.renderArea().bbox();
>         return bbox.center().sub(bbox.pos);
>     }
>     else {
>         return null
>     }
> }
> ```

I don't see how that wouldn't work for text, it's just the format width / 2, same with height.
But the center methods just gets the center based off of half the specified size.

```ts
const a = rect(300, 250)
const c = a.center() // -> vec2(300 / 2, 250 / 2)
```

mflerackers:
I think it is better to get that center from the renderArea, since all draw components support renderArea.

mflerackers:
I also still have no idea what the significance or use case for this method is if it would be included. I can see a use case for the actual center relative to the object origin, but not one for the center relative to the top-left of the shape.

SpcFORK:
> I think it is better to get that center from the renderArea, since all draw components support renderArea.

image
image

As you can see here, I use the parameters instead without a vector to make the widths and height of this UI.
If there were a center method on the RectComp, I could just use `.add` instead.

Currently width and height are just loose ints that sit in the GameObj, and no draw object contains a center method for bundling the width and height they are passed.

mflerackers:
But all draw objects have object.renderArea().bbox() from which you can get width, height and center.
What you call center is actually halfSize, halfWidth and halfHeight, it is not center.

SpcFORK:
> But all draw objects have object.renderArea().bbox() from which you can get width, height and center. What you call center is actually halfSize, halfWidth and halfHeight, it is not center.

??

halfWidth and halfHeight will always get the center of a rect, disregarding any anchor?
And radius would do the same with a circle?

mflerackers:
The halfWidth and halfHeight of an object have no relation with its position. They are

const halfWidth = object.renderArea().bbox().width / 2;
const halfHeight = object.renderArea().bbox().height / 2;

The center of an object however is relative to a specific coordinate system

object.renderArea().bbox().center() is the center relative to the object
object.renderArea().transform(object.transform).bbox().center() is the center relative to the world
object.renderArea().transform(object.transform)transform(game.cam.transform).bbox().center() is the center relative to the camera

SpcFORK:
halfWidth and halfHeight are essential though to making UI since you wouldn't want to do it globally to make it work well
And after transform, should be relative.

Using vectors rather than math is cleaner and less gross

SpcFORK:
If I am not mistaken, renderArea() also creates a new Rect as well.
AND THEN bbox does clone

mflerackers:
That's why I called it halfSize, which would be a vector containing halfWidth and halfHeight.

But we can't just add some functions to some components and not others. Like now it would only work for circles, sprites and rect. It would be no better than the width and height we have now. There is no interface defined or general rule of which components should implement that method. It is already not great that draw objects arbitrarily have a renderArea method in order to make area work.

Adding a function is not something that should be done without thinking about use cases and specifications, as once it is added, you can't just rename or remove it without breaking people's games. It may be that there is a valid use case for this to be added, but the naming, which components should implement it, whether there is a formal interface etc should be decided.

For example, it might be best to make a DrawableComp interface (or LayoutComp from which DrawableComp derives so that hbox, vbox or grid can implement layout methods even though they don't draw) defining the renderArea and other layout related methods. Then make SpriteComp, RectComp, etc derive from DrawableComp instead of Comp. This way is cleaner than just adding a convenience method to some components because you need them at the moment.

niceEli:
there is conflicts now

PR #255, opened 2024-07-13
refactor: move game and make to core, full context

lajbel:
This separates the game variable in `initGame` and separate `make()` function inside the core folder


mflerackers (approved):
Maybe we need some benchmarks to see the possible difference in performance?

lajbel:
I compared with v3000, they looks the same, maybe there's a way to create a more accurate benchmark?

PR #266, opened 2024-07-16
feat: type `on()`

lajbel:
## Description

Types all `on()` events that are in the source related to components and raw game obj


mflerackers (approved):
👍

PR #274, opened 2024-07-20
Fix easing demo

luizbills:
## Description

Easing demo is broken in https://play.kaplayjs.com/?example=easing

### DIFF
--- a/examples/easing.js
+++ b/examples/easing.js
@@ -10,7 +10,7 @@ add([
 ]);
 
 onClick("steps", (square) => {
-    this.tween(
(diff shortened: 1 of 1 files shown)


mflerackers (changes requested):
It should be square.tween, not tween. Global tween use should be discouraged as it isn't safe.

luizbills:
> It should be square.tween, not tween. Global tween use should be discouraged as it isn't safe.

Make sense.

Done!

mflerackers (approved):

PR #280, opened 2024-07-25
chore: fix typo in add example

Ahmad3366:
it should be called `kaplay` instead of `kaboom` now, right ?


lajbel (approved):

lajbel:
A little edition by me, good catch!

PR #290, opened 2024-07-28
fix: allow the same input button inside different input bindings

### LINKED ISSUE #286: feat: multiple virtual buttons associated with physical input key
I am making a platformer game that allows the player to jump *and* climb ladders. The controls associated with each have overlaps -- for example you can use space, w, or up to jump when walking normally not on a ladder, or you can use w or up to climb upwards when on a ladder, and the state (on ladder or off ladder) can be manually toggled by the player as well with another button.

I na&iuml;vely thought that just adding the keybindings in the global kaplay options would work, so I wrote this:

```js
kaplay({
    // snip
    buttons: {
        jump: {
            keyboard: ["w", "up", "space"],
            gamepad: ["south"]
        },
        move_up: {
            keyboard: ["w", "up"],
        },
        // snip
    }
});
```

... and then I wrote some code and wondered "why isn't my onButtonDown handler for `move_up` ever being called?"

The reason is that Kaplay just stops the first virtual button that includes the physical button that was just pressed, and doesn't bother to look further.

## **Describe the solution you'd like**

Simple: anywhere the functions that look up what virtual button a particular input event is associated with are called, they can be changed so they return an array of matching buttons instead of just the first one:

https://github.com/kaplayjs/kaplay/blob/c6ca658a03a86bc58af2f2b89b2b4f07156e9e00/src/app/app.ts#L943-L950

So the return type of `getButtonNameByKey` is currently `string | undefined`, perhaps it can be changed to `string[]`, and then line 947 can be replaced with something like this:
```ts
getButtonNamesByKey(k).forEach(btn => state.buttonState.press(btn));
```

## **Describe alternatives you've considered**

I don't need to or want to use a scene to do this switching - it's literally just one global boolean variable that some event handlers look at.

lajbel:
- Related Issue #
- Closes #286

### DIFF
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -25,6 +25,7 @@ import {
     type ButtonBinding,
     type ButtonsDef,
     getLastInputDeviceType,
+    parseButtonBindings,
 } from "./inputBindings";
 
 export class ButtonState<T = string> {
@@ -87,9 +88,14 @@ export const initAppState = (opt: {
     maxFPS?: number;
     buttons?: ButtonsDef;
 }) => {
+    const buttons = opt.buttons ?? {};
+
     return {
         canvas: opt.canvas,
-        buttons: opt.buttons ?? {},
+        buttons: buttons,
+        buttonsByKey: new Map<Key, string[]>(),
+        buttonsByMouse: new Map<MouseButton, string[]>(),
+        buttonsByGamepad: new Map<KGamepadButton, string[]>(),
         loopID: null as null | number,
         stopped: false,
         dt: 0,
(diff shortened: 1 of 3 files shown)


niceEli (approved):

dragoncoder047:
how can I just link to the built dist file for this commit? I couldn't find anything on the pkg.pr.new documentation

lajbel:
Install the package, then go to `dist/kaplay.js` or use "node_modules/kaplay/dist/kaplay.js"

dragoncoder047 (approved):

dragoncoder047:
It works good!!

PR #301, opened 2024-08-02
feat: new `body()` events, `land` and `headbutted`, mirrors of `headb…

### LINKED ISSUE #300: feat: event when an object is stomped on or headbutted (reflected forms of `onGround` and `onHeadbutt`)
There's events for when an object hits the floor (`onGround`) and hits the ceiling (`onHeadbutt`). Where are the equivalents for the thing that was hit (i.e. `onStomped` and `onHeadbutted`)?

I'm making a button object that will react when it a falling object (which may or may not be the player) lands on it from above, and for the event handler that makes the stomped-on button react to the stomp, it doesn't make much sense to attach it to the thing that's falling (by tag or otherwise)...

lajbel:
- Closes #300

### DIFF
--- a/src/components/physics/body.ts
+++ b/src/components/physics/body.ts
@@ -122,6 +122,16 @@ export interface BodyComp extends Comp {
      * @since v2000.1
      */
     onHeadbutt(action: () => void): KEventController;
+    /**
+     * Register an event that runs when an object lands on this object.
+     *
+     * @since v3001.0
+     */
+    onLand(action: (obj: GameObj) => void): KEventController;
+    /**
+     * Register an event that runs when the object is bumped by another object head.
+     */
+    onHeadbutted(action: (obj: GameObj) => void): KEventController;
 }
 
 /**
@@ -258,6 +268,7 @@ export function body(opt: BodyCompOpt = {}): BodyComp {
                             }
                             else {
                                 this.trigger("ground", curPlatform);
+                                col.target.trigger("land", this);
                             }
                         }
                         else if (col.isTop() && this.isJumping()) {
@@ -265,6 +276,7 @@ export function body(opt: BodyCompOpt = {}): BodyComp {
                                 game.gravity.unit(),
                             );
                             this.trigger("headbutt", col.target);
+                            col.target.trigger("headbutted", this);
                         }
                     }
                 });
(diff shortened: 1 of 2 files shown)


dragoncoder047 (commented):
would've done this change myself but I'm working on the rest of the game :)

dragoncoder047:
also, maybe a separate issue/pr, but can an event when the object that triggered the "land" event (and itself got a "ground" event) moves away, can the object get an "unLand" event? I am calling it "unstomp" but you get the idea. Not sure exactly what the un-headbutted event should be called.

lajbel:
when an object is out of a platformer, it calls onCollideEnd, the same with the stomp

niceEli (approved):

dragoncoder047:
> when an object is out of a platform, it calls onCollideEnd, the same with the headbutt

Ok, this works, but I noticed that if the player clips the corner of the button as you're jumping up on to the button, it triggers the "onCollideEnd" too, but it wouldn't be a valid "unstomped" event. How would I check if it's a valid "unstomp"?

lajbel:
can you give me a playground link?

dragoncoder047:
> can you give me a playground link?

The kaplayground doesn't have the latest version with `onLand()` yet, but this should work (when it does):

```js
kaplay();
loadSprite("bean", "/sprites/bean.png");
loadSprite("spike", "/sprites/spike.png");
loadSprite("grass", "/sprites/grass.png");

const SPEED = 480;

setGravity(2400);

const level = addLevel([
    "@  ^   ",
    "=======",
], {
    tileWidth: 64,
    tileHeight: 64,
    pos: vec2(100, 200),
    tiles: {
        "@": () => [sprite("bean"), area(), body(), anchor("bot"), "player"],
        "=": () => [sprite("grass"), area(), body({ isStatic: true }), anchor("bot")],
        "^": () => [sprite("spike"), area(), body({ isStatic: true }), anchor("bot"), "danger"],
    },
});

const player = level.get("player")[0];
const spikes = level.get("danger")[0];

onKeyPress("space", () => {
    if (player.isGrounded()) {
        player.jump();
    }
});

onKeyDown("left", () => {
    player.move(-SPEED, 0);
});

onKeyDown("right", () => {
    player.move(SPEED, 0);
});

spikes.onCollideEnd(() => {
    debug.log("unstomped");
});

spikes.onLand(() => {
    debug.log("stomped");
});
```

Also, in this example the "unstomped" log occurs when you bonk the spikes from the side, and then move away (as well as the corner clipping I described above: hold arrow key to go against the spikes, while still holding the key, press space). So there needs to be more filtering on whether it's a valid "unstomp".

PR #302, opened 2024-08-02
fix: add error when error is not an error

### LINKED ISSUE #299: bug: error screen errors out when error is not an Error
...and yes, I did put "error" into the title four times!

# Describe the bug

Some event handlers have the feature that when they throw a Javascript error, the game halts with a big blue screen showing the error message.

However, I noticed (by accident) that if the thrown error isn't actually an instance of the Javascript Error class (I occasionally insert `throw "blooey!"` into my code to stop it prematurely), then the code that attempts to display the error itself throws an error and the game halts but to a blank black screen. The error from the dev tools console is `Error: formatText() requires property "text".` and it keeps getting spammed in a loop.

## Version

Whatever's in the kaplayground. Also, bleeding edge master version (`npm i https://github.com/kaplayjs/kaplay`)
## To Reproduce

Put this in kaplayground. Open browser dev tools console. Set `notErrorError` to false to see the normal error screen, or to true to see the brokenness if a string is thrown.
```js
const notErrorError = true;

kaplay();
add([pos(10, 10), rect(10, 10), body(), area(), color("#ff0000"), "A"]);
add([pos(10, 30), rect(10, 10), body({ isStatic: true }), area(), color("#ffff00"), "B"]);
setGravity(200);
onCollide("A", "B", () => {
    if (notErrorError) throw "blooey!" // error 1: not an Error object
    else kaplay.foo(); // error 2: is an Error object
});
```

## Expected behavior

The normal blue error screen displays in both cases with the error message.

## Screenshots

with notErrorError = false:
Screenshot 2024-08-02 at 3 35 33 PM

with notErrorError = true:
Screenshot 2024-08-02 at 3 36 23 PM

lajbel:
image

- Close #299

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -148,6 +148,7 @@ obj.sprite = "bag";
 
 - **(break)** much typescript definitions was fixed, if you use typescript now
   maybe you see new errors that make your code strict
+- fix error screen not showing with not Error object
 
 ## Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,7 +129,7 @@ obj.sprite = "bag";
 - added `navigation()` component to calculate a list of waypoints on a graph
 - added `animate()` component to animate the properties of an object using
   keyframes
-- added effector components: `areaEffector()`, `buoyancyEffector`,
+- added effector components: `areaEffector()`, `buoyancyEffector()`,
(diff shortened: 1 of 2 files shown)


dragoncoder047:
I would've just done `err.toString()` as a default to support throwing strings like this.

mflerackers (approved):

PR #316, opened 2024-08-04
feat: inspect properties of objects

### LINKED ISSUE #17: feat: A way to be able to see the properties a GameObj has on debug.inspect
### Discussed in https://github.com/marklovers/kaplay/discussions/16

<div type='discussions-op-text'>

<sup>Originally posted by **amyspark-ng** May 22, 2024</sup>
Let's say my obj has a property speed, that i access doing obj.speed, i'd like to be able to see it in the debug.inspect window when you hover on it, not only the properties Comps return but properties that i myself have added</div>

lajbel:
- close #17


mflerackers (commented):
Shouldn't we check if the anonymous comp has an inspect function?

lajbel:
True, will do

PR #331, opened 2024-08-07
feat: add panner node into play() pipeline

### LINKED ISSUE #304: feat: Stereo pan option for `play()`
We already have a volume option for `play()`ing a sound. Can stereo pan be added? There is a web audio node to handle this, it just needs to be put in the pipeline.

This would be useful when a sound is triggered by some physical action, and it would make sense to have the sound appear to come from the location of that action. For example, if an enemy spawned offscreen to the left, the player could be cued that it is to the left by the spawn sound coming from the left speaker. (This is particularly evident when you are using headphones.)

dragoncoder047:
closes #304

### DIFF
--- a/src/audio/play.ts
+++ b/src/audio/play.ts
@@ -280,7 +280,7 @@ export function play(
 
         set pan(pan: number) {
             panNode.pan.value = pan;
-        }
+        },
 
         get pan() {
             return panNode.pan.value;
--- a/src/audio/play.ts
+++ b/src/audio/play.ts
@@ -43,6 +43,12 @@ export interface AudioPlayOpt {
      * The start time, in seconds.
      */
     seek?: number;
+    /**
+     * The stereo pan of the sound.
+     * -1.0 means fully from the left channel, 0.0 means centered, 1.0 means fully right.
+     * Defaults to 0.0.
+     */
+    pan?: number;
 }
 
 export interface AudioPlay {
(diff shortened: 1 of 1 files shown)


lajbel:
why is gainNode not used anymore?

dragoncoder047:
No, is still used, source connects to pan node, pan node connects to gain node, gain connects to output.

lajbel (approved):

mflerackers (approved):

lajbel:
you should open a PR with the same changes but for master

dragoncoder047:
> you should open a PR with the same changes but for master

I did originally base this pr against master, you switched it!

lajbel:
IK, what I mean is create another PR because it's a change compatible with v4000 and v3001

lajbel:
This should be only for v3001, shouldn't merge with master

lajbel:
is on https://github.com/kaplayjs/kaplay/blob/master/CONTRIBUTING.md

niceEli (approved):

niceEli:
the error is not fault of the pr. its a pkg.pr.new issue

lajbel:
> > IK, what I mean is create another PR because it's a change compatible with v4000 and v3001
> 
> so you mean merge this into v3001, and then do the same changes but on master?

You should replicate the same changes for every branch, master and v3001, then open 2 PR's

dragoncoder047:
> You should replicate the same changes for every branch, master and v3001, then open 2 PR's

Ok, I switched this one back to master to keep the commit history clean, will open another pr for v3001 soon

niceEli:
is ready for merge?

niceEli:
btw you dont really need to worry about commit history as it just gets squashed and merged into 1 commit anyway

PR #370, opened 2024-08-29
feat: add keyboardCode on input bindings

### LINKED ISSUE #368: feat: When processing input, get the ID (code) of the physical button on the keyboard
For universal input. If there are two languages in the operating system, then when processing, I need to process all the characters (all that can be on the keyboard). There is no clear link to the "physical" button ([code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)).
_Example_: I am processing the input of the `W` character, which is passed to key: Key, but if I switch to a different layout (another language), then when I click on the button, I will get a different character. You need to keep an eye on the layout when playing.

Solution options:
1. In the input handler (OnKeyDown, OnKeyPress and others), add a new argument in which the code will be output.

```typescript
onKeyDown((key: Key, code: string) => {
  if (code === "keyZ") { // So this is true for both when key = z or key = w (with AZERTY keyboard layout)
    // TODO
  }
})
```
2. Expand the `Key` type by adding new values.
```typescript
onKeyDown((key: Key) => {
  if (key.data === "w") {
    // TODO
  }
  if (key.code === "keyZ") { // So this is true for both when key = z or key = w (with AZERTY keyboard layout)
    // TODO
  }
})
```

lajbel:
New syntax:

```js
kaplay({
    buttons: {
        "jump": {
            "keyboardCode": ["Period", "Space"],
        }
    }
})
```

- Closes #368

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ features of v4000, while v4000 will have the most features and breaking changes.
       buttons: {
           jump: {
               keyboard: ["space", "up"],
+              keyboardCode: "Space", // you can also use key codes
               gamepad: ["south"],
           },
       },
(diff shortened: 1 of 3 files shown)


lajbel:
@e1e5en-gd how it looks for you?

e1e5en-gd:
> @e1e5en-gd how it looks for you?

Good

mflerackers (approved):

PR #372, opened 2024-08-30
feat: fake mouse

### LINKED ISSUE #359: feat: make "fake cursor" still trigger hover events
### Discussed in https://github.com/kaplayjs/kaplay/discussions/330

<div type='discussions-op-text'>

<sup>Originally posted by **dragoncoder047** August  6, 2024</sup>
So I decided that my game would not force people to use their mouse. To do that I requested pointer lock (i.e. `setCursorLocked(true)`), added a game object with a custom cursor sprite and a z-index of infinity, and programmed it to move when the mouse moves, but also move according to the right gamepad stick.

Unfortunately with pointer lock turned on Kaplay no longer knows where the pointer is and so the area component can't fire its `onHover` event.

How do I tell Kaplay that my synthesized pointer should be used for the hover events and/or feed the synthesized pointer's position into the mouse-hover logic?</div>

lajbel:
- Closes #359

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,7 +70,7 @@ features of v4000, while v4000 will have the most features and breaking changes.
 
   ```js
   const myCursor = add([
-      fakeCursor(),
+      fakeMouse(),
       sprite("kat"),
       pos(100, 100),
   ]);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,20 @@ features of v4000, while v4000 will have the most features and breaking changes.
   });
   ```
 
+- added a fake cursor API (**v4000**)
+
+  ```js
+  const myCursor = add([
+      fakeCursor(),
+      sprite("kat"),
+      pos(100, 100),
+  ]);
+
+  myCursor.press(); // trigger onClick events if the mouse is over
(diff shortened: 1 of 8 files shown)


lajbel:
@dragoncoder047 how it looks for you?

CHANGELOG.md
```diff
@@ -66,6 +66,20 @@ features of v4000, while v4000 will have the most features and breaking changes.
   });
   ```
 
+- added a fake cursor API (**v4000**)
+
+  ```js
+  const myCursor = add([
+      fakeCursor(),
```
dragoncoder047: shouldn't this be `fakeMouse()` ?

lajbel: true, fixed

niceEli (approved):

dragoncoder047:
just should note that if the camera is transformed with `camPos()` the cursor will be *drawn* in the wrong position but everything else will be fine. E.g. add `camPos(vec2(100, 100))` to the bottom of the new fakeMouse example to see this.

Adding `fixed()` to the mouse object fixes it, might want to make it a dependency component of fakeMouse, or just make a note in the documentation?

lajbel:
I think add required component is fine

PR #383, opened 2024-09-02
feat: connect play() audio to somewhere other than default node

dragoncoder047:
allows the user to redirect the audio from play() to a different audio node other than Kaplay's global gain node


lajbel (approved):
Good changes :) Thank you!

Also, the correct form to spell KAPLAY is KAPLAY, uppercase, would you change it

PR #385, opened 2024-09-03
swap translate <--> scale transforms

### LINKED ISSUE #381: bug: areas stretched with scale, debug outline doesn't line up with real area used for collisions and inspect hover
Now that #338 is "closed", the debug areas are drawn not-stretched, except they don't actually line up with the real area.

Here is a demo that not only shows that they don't line up, but also is what I was doing when I discovered both #338 and this bug.

Paste in kaplayground:

```js
const doMergeAcross = false;
kaplay({ scale: 0.5 });
loadBean();
loadSprite("grass", "/sprites/grass.png");
const SPEED = 480;
setGravity(2400);
const level = addLevel([
    "  @",
    "=============",
], {
    tileWidth: 64,
    tileHeight: 64,
    pos: vec2(100, 200),
    tiles: {
        "@": () => [
            sprite("bean"),
            area(),
            body(),
            anchor("center"),
            "player",
        ],
        "=": () => [
            sprite("grass"),
            area(),
            body({ isStatic: true }),
            anchor("center"),
            "wall",
        ],
    },
});
const player = level.get("player")[0];
onKeyPress("space", () => {
    if (player.isGrounded()) {
        player.jump();
    }
});
onKeyDown("left", () => {
    player.move(-SPEED, 0);
});
onKeyDown("right", () => {
    player.move(SPEED, 0);
});

if (doMergeAcross) mergeAcross();
debug.inspect = true;

// merge across:

function mergeAcross() {
    const tw = level.tileWidth();
    for (var y = 0; y < level.numRows(); y++) {
        var prevTile = null;
        for (var x = 0; x < level.numColumns(); x++) {
            const pos = vec2(x, y);
            const thisTile = level.getAt(pos)[0];
            if (
                prevTile != null
                && prevTile.is("wall")
                && thisTile != null
                && thisTile.is("wall")) {
                prevTile.area.offset.x += tw / 2;
                prevTile.area.scale.x++;
                thisTile.unuse("area");
                thisTile.unuse("wall");
            }
            else {
                if (thisTile != null && thisTile.is("wall"))
                    prevTile = thisTile;
                else
                    prevTile = null;
            }
        }
    }
}
```

With doMergeAcross set to false, everything is as expected, except there are multiple wall tiles. I don't want this.

Sooooo... with doMergeAcross set to true, some bad things happen:

1. The blue box, that is supposed to be the area, lines up perfectly with the outline of all the wall tiles. BUT...
2. The player falls right through the floor.
3. And hovering over the blue box, no inspector pop up appears, indicating that *you're not actually hovering on any area*.

why isn't the debug view lining up when the area scale is not (1, 1)?

dragoncoder047:
ref: https://github.com/kaplayjs/kaplay/issues/381#issuecomment-2324826289

@mflerackers : "translate and scale transforms were done in the wrong order before drawing for some reason. It's an easy fix..."

is this the fix?

closes #381 if it is

### DIFF
--- a/src/components/physics/area.ts
+++ b/src/components/physics/area.ts
@@ -531,8 +531,8 @@ export function area(opt: AreaCompOpt = {}): AreaComp {
 
             const transform = this.transform
                 .clone()
-                .scale(vec2(this.area.scale ?? 1))
-                .translate(this.area.offset);
+                .translate(this.area.offset)
+                .scale(vec2(this.area.scale ?? 1));
 
             if (localArea instanceof k.Rect) {
                 const offset = anchorPt(this.anchor || DEF_ANCHOR)
--- a/src/kaplay.ts
+++ b/src/kaplay.ts
@@ -764,8 +764,8 @@ const kaplay = <
             stack.push(tr.clone());
 
             // Update object transform here. This will be the transform later used in rendering.
-            if (obj.scale) tr.scale(obj.scale);
             if (obj.pos) tr.translate(obj.pos);
+            if (obj.scale) tr.scale(obj.scale);
             if (obj.angle) tr.rotate(obj.angle);
             obj.transform = tr.clone();
 
--- a/src/kaplay.ts
+++ b/src/kaplay.ts
@@ -764,8 +764,8 @@ const kaplay = <
             stack.push(tr.clone());
 
             // Update object transform here. This will be the transform later used in rendering.
-            if (obj.pos) tr.translate(obj.pos);
             if (obj.scale) tr.scale(obj.scale);
+            if (obj.pos) tr.translate(obj.pos);
             if (obj.angle) tr.rotate(obj.angle);
             obj.transform = tr.clone();
 


mflerackers:
No, certainly not. That would completely break collision detection itself.

dragoncoder047:
> No, certainly not. That would completely break collision detection itself.

So, where is the "out-of-order" transformations at?

dragoncoder047:
This PR is empty now, but if I find the out-of order transformations I'll fix it here

mflerackers:
Possibly here, applying the scale is done before offset. But you shouldn't only change the code, but test it as well.

worldArea(this: GameObj<AreaComp | AnchorComp>): Polygon {
            const localArea = this.localArea();

            const transform = this.transform
                .clone()
                .scale(vec2(this.area.scale ?? 1))
                .translate(this.area.offset);

dragoncoder047:
ok I tried it there

how do I re-trigger the pkg.pr.new thing so I can so I can just `npm install` my changes to test them?

mflerackers:
Didn't it update itself? I think it did.

dragoncoder047:
> Didn't it update itself? I think it did.

Yeah it did (I noticed like 30 seconds after I added that comment, see the edit)

> But you shouldn't only change the code, but test it as well.

I installed it and tested, it seems to work for me.

mflerackers (approved):
👍

PR #395, opened 2024-09-05
update velocity when moving if available

dragoncoder047:
I was trying to write some code that used the player's velocity to make footstep noises when the player moves, but the pos component's move function just updates the position, not the velocity. I had some horribly ugly hacks in place to get it working.

Why not add the move()'s "simulated velocity" to the real velocity to make it work instead? Then you can check the velocity to get the the actual motion of the object.

(I will test this and close the PR if it obviously doesn't work.)

### DIFF
--- a/src/components/transform/pos.ts
+++ b/src/components/transform/pos.ts
@@ -83,7 +83,7 @@ export function pos(...args: Vec2Args): PosComp {
             const vel = vec2(...args).scale(k.dt());
             this.moveBy(vel);
             if (this.c("body")) {
-                this.vel = vel;
+                this.vel.add(vel);
             }
         },
 
--- a/src/components/transform/pos.ts
+++ b/src/components/transform/pos.ts
@@ -80,7 +80,11 @@ export function pos(...args: Vec2Args): PosComp {
 
         // move with velocity (pixels per second)
         move(...args: Vec2Args) {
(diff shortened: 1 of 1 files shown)


dragoncoder047:
OK, that's very weird. I tried it and it *COMPLETELY* prevents movement; and additionally all of the fixed bodies fail to render at all (Why?)

dragoncoder047:
> Shouldn't you then just use the `addForce()` or `applyImpulse()` available on `BodyComp` for movement?

It's not exactly equivalent, since if I use move() the body stops moving instantly when I stop calling move(), but the if I use addForce(), it will continue to slide until it hits something or slows down enough.

imaginarny:
Would it be possible to reset the velocity in that case to stop?

mflerackers:
> > Shouldn't you then just use the `addForce()` or `applyImpulse()` available on `BodyComp` for movement?
> 
> It's not exactly equivalent, since if I use move() the body stops moving instantly when I stop calling move(), but the if I use addForce(), it will continue to slide until it hits something or slows down enough.

You need to take into account the current velocity when calculating your force. Like the surface effector does for example.

mflerackers:
BTW, the vector rejection of a vector on the same vector is 0 because there is no orthogonal component.

PR #426, opened 2024-10-03
closest orthagonal distance function

dragoncoder047:
idk this might be useful?

### DIFF
--- a/src/math/math.ts
+++ b/src/math/math.ts
@@ -112,11 +112,11 @@ export class Vec2 {
     static UP = new Vec2(0, -1);
     static DOWN = new Vec2(0, 1);
 
-    /** Closest orthagonal direction: LEFT, RIGHT, UP, or DOWN */
+    /** Closest orthogonal direction: LEFT, RIGHT, UP, or DOWN */
     toAxis(): Vec2 {
         var result = vec2();
-        var minDist = Number.MAX_VALUE;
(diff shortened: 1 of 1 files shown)


lajbel (approved):
LGTM!

imaginarny:
Wait a second. There are a few cosmetic issues. It is 'orth**O**gonal', not 'orth**A**gonal'. Also, `let` would be preferable instead of `var`, although it's not much of an issue in this particular case (probably). Just a heads-up.

mflerackers:
Isn't this just return vec2(this.x < 0 ? left : right, this.y < 0 ? up : down) ?

dragoncoder047:
> Isn't this just return vec2(this.x < 0 ? left : right, this.y < 0 ? up : down) ?

Not quite. It does the orthogonal direction closest to the angle of this, so say if you have vec2(-1, -10) it will return UP not LEFT.

Since it is only 4 candidates, I suspect it can be done with a bunch of conditionals, however, it made more sense to use a loop to me. :shrug: It might be *faster* to use conditionals, but until this function is used somewhere I really can't say.

mflerackers:
So 

return Math.abs(this.x) > Math.abs(this.y) ? 
  this.x < 0 ? left : right :
  this.y < 0 ? up : down;

Two conditionals will always be faster than calculating 4 square roots (which weren't really needed, the dot product is enough to compare distances).

mflerackers (approved):

PR #436, opened 2024-10-06
fix: fromScreen not using viewport scale, mousePos() not working with letterbox/stretch

### LINKED ISSUE #325: bug: fromScreen() and friends ignore global scale
put in kaplayground:

```js
kaplay({
    scale: 3
});

const x = add([
    anchor("center"),
    circle(3),
    color(RED),
    pos(),
]);

getTreeRoot().use(pos());

onMouseMove(pos => {
    x.pos = getTreeRoot().fromScreen(pos);
});
```
red circle should be under cursor but it is not (it is exactly 3x farther away from (0, 0) than the mouse because scale is 3)

> Added by lajbel

- [x] Fix in v3001
- [x] Fix in v4000

lajbel:
### Issues or related

- Related Issue #
- Closes #325


mflerackers (changes requested):
This is wrong. World and local have nothing to do with viewport scale.
Viewport scale only makes sense between screen and world.

mflerackers:
Additionally, you don't want to divide each time, it is better to store 1 / factor and multiply instead.

dragoncoder047:
So scale is fine now, it's stretch that is bugged. If `letterbox: true` then it works since the canvas's aspect ratio is as specified. `stretch` makes it not be as specified, so it gets off.

```js
kaplay({
    scale: 3,
    width: 20,
    height: 40,
    stretch: true
});

const x = add([
    anchor("center"),
    circle(3),
    color(RED),
    pos(),
]);

getTreeRoot().use(pos());

onMouseMove(pos => {
    x.pos = getTreeRoot().fromScreen(pos);
});
```

mflerackers:
Yeah, when stretch is used, you no longer have a uniform scale. You need a vec2 in that case.

lajbel:
> So scale is fine now, it's stretch that is bugged. If `letterbox: true` then it works since the canvas's aspect ratio is as specified. `stretch` makes it not be as specified, so it gets off.
> 
> ```js
> kaplay({
>     scale: 3,
>     width: 20,
>     height: 40,
>     stretch: true
> });
> 
> const x = add([
>     anchor("center"),
>     circle(3),
>     color(RED),
>     pos(),
> ]);
> 
> getTreeRoot().use(pos());
> 
> onMouseMove(pos => {
>     x.pos = getTreeRoot().fromScreen(pos);
> });
> ```

The question is if there's an user case using only Stretch without letterbox

niceEli:
resolved conflicts in commit 894d9df6d65baa497d017fb09db200cb39066eae

niceEli:
no idea for fail
should of worked properly

PR #443, opened 2024-10-09
fix: add return if static to physics apply force functions

dragoncoder047:
I noticed by accident that if a static body collides with an area effector, then the velocity of the static body will still be modified and the static body will fly off into the ether. But it's static, it shouldn't move. This fixes that.

```ts
kaplay();

add([
    pos(10, 10),
    rect(100, 100),
    color(BLUE),
    area(),
    areaEffector({ forceAngle: 0, forceMagnitude: 10 }),
]);

// Wheeeeeee!! I am static but still get pushed!
add([
    pos(20, 20),
    rect(100, 10),
    color(RED),
    area(),
    body({ isStatic: true }),
]);
```


niceEli (changes requested):
maybe there could be an error like "cannot add a force too a static body" or something like that, other than that it seems good, if you feel it shouldnt have an error then just say and ill approve

niceEli:
wait sorry nvm to that i reread what you wrote

niceEli (approved):

mflerackers (approved):

lajbel:
This is both versions? We could say it was a bug?

PR #455, opened 2024-10-14
fix: platformEffector shouldCollide this argument

dragoncoder047:
The shouldCollide function of the platform effector is now called with the game object as the "this" argument, so it can refer to its own properties.


niceEli (approved):
seems like a good simple change

PR #457, opened 2024-10-15
docs: fix spelling mistakes in jsdoc (+ a bit of formatting)

ErikGXDev:
Fixed spelling mistakes in types.ts, it seems that some parts of the code have been formatted differently after running `dprint fmt`

PR #458, opened 2024-10-15
docs: fix spelling mistakes in jsdoc (for v3001)

ErikGXDev:
Basically identical to #457 but for the v3001 branch.


ErikGXDev:
Oops, forgot that I should only PR to the master branch

PR #459, opened 2024-10-15
feat: added small yet significant changes for particles()

tristanperrow:
## Description

1. The variables `direction` and `spread` can now be set after object creation with `setDirection(newDirection)` and `setSpread(newSpread)`. This way you can achieve particle effects you couldn't beforehand.
2. The `scale` of the particles now are correctly lerped.
3. Added `angleModifier` type & parameter to `ParticlesOpt` so you can make it so the angle of the particle is defined by the direction it is moving.

## Other notes

- Did `pnpm run fmt` and there were some files that weren't previously formatted.
- Also totally open for API changes, because this is just a way to make this work. If we have any ideas on how to improve the API here, please share :).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,14 @@
   });
   ```
 
+- added `pressButton(btn)` and `releaseButton(btn)` to simulate button press and
+  release
+
+  ```js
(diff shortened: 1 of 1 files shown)


mflerackers (changes requested):
Isn't it better to use get/set properties instead of functions?

PR #466, opened 2024-10-19
docs: add some examples to some comps

amyspark-ng:
yea


src/types.ts
```diff
@@ -368,7 +368,18 @@ export interface KAPLAYCtx<
      *
      * @param a The angle to rotate by. Defaults to 0.
```
dragoncoder047, line 371: might be worthwhile to note that angles for this are in degrees, some other game libs use radians

lajbel: true

amyspark-ng: doesn't it already say (in degrees)

dragoncoder047: oh duh, it was the line that got cut off by the GitHub diff

amyspark-ng:
if it's good then push pls

niceEli:
conflicts fixed

niceEli (approved):

PR #474, opened 2024-10-22
jsdoc example for onadd

amyspark-ng:
before submitting a pull request

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1282,10 +1282,10 @@ export interface KAPLAYCtx<
      * ])
      * 
      * bean.onAdd(() => {
-     *     bean.pos = center()
+     *     bean.pos = center();
      * })
      * 
-     * add(bean)
+     * add(bean);
      * 
      * @group Events
      */
@@ -4290,10 +4290,10 @@ export interface GameObjRaw {
      * ])
      * 
      * bean.onAdd(() => {
-     *     bean.pos = center()
+     *     bean.pos = center();
      * })
      * 
-     * add(bean)
+     * add(bean);
      * 
      * @group Events
      */
--- a/src/types.ts
+++ b/src/types.ts
@@ -1263,10 +1263,30 @@ export interface KAPLAYCtx<
      */
     onDraw(action: () => void): KEventController;
     /**
+     * Runs when an object with a certain tag is added to the scene.
+     * 
+     * ```js
+     * onAdd("coin", (coin) => {
+     *     play("bling")
+     * })
+     * ```
      * @group Events
      */
     onAdd(tag: Tag, action: (obj: GameObj) => void): KEventController;
     /**
+     * Runs when an specific object is added to the scene.
+     * 
+     * let bean = make([
+     *     sprite("bean"),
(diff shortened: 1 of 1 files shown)


imaginarny:
There shouldn't be the same example for top level `onAdd` and `gameObjRaw` method `onAdd`. You are missing use cases on top level one as we discussed on Discord.

lajbel:
You forgot some ;

amyspark-ng:
> There shouldn't be the same example for top level `onAdd` and `gameObjRaw` method `onAdd`. You are missing use cases on top level one as we discussed on Discord.

Could you better explain? so that i can put i was checking discord and i really don't remember what we talked about

src/types.ts
```diff
@@ -1263,10 +1263,30 @@ export interface KAPLAYCtx<
      */
     onDraw(action: () => void): KEventController;
     /**
+     * Runs when an object with a certain tag is added to the scene.
+     * 
+     * ```js
+     * onAdd("coin", (coin) => {
+     *     play("bling")
```
niceEli: pls add semicolons

niceEli (changes requested):
semicolons, then its perfect

imaginarny:
> Could you better explain? so that i can put i was checking discord and i really don't remember what we talked about
— @amyspark-ng 

There is a top level `onAdd` and `onAdd` that is a method of `gameObjRaw`. Top level can either accept a tag and action as args, or just an action. You have an example for a tag option, which is ok. But [the next one, for the action only](https://github.com/kaplayjs/kaplay/pull/474/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR1277), should be something like this:

```js
// Runs when an object is added to the scene.

onAdd((obj) => {
    obj.pos = obj.sprite == "bean" ? center() : vec2(rand(0, width()), rand(0, height()));
})

add([sprite("bean")]); // centered pos
add([sprite("bobo")]); // random pos
```

Instead of what you have:

```js
// Runs when an specific object is added to the scene.

let bean = make([
    sprite("bean"),
    pos(),
])

bean.onAdd(() => {
    bean.pos = center();
})

add(bean);
 ```

Point is that it's not called on an object, instead uses `obj: GameObj` param from action. Also, notice the difference in the description. First runs when an object (any) is added to the scene, not a specific one as in the case for `gameObjRaw` (there is a typo too, should be ~an~ a specific).

Then you have this example 2nd time also for `gameObjRaw` [there](https://github.com/kaplayjs/kaplay/pull/474/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR4285), which is right.

Hope it makes more sense now.

amyspark-ng:
hopefully this is good now

amyspark-ng:
WAIT don't merge this, https://github.com/kaplayjs/kaplay/pull/477 is way better

amyspark-ng:
closed for https://github.com/kaplayjs/kaplay/pull/477, which is 10 times better

PR #475, opened 2024-10-23
fix: brackets causing issues in debug.log + enhance logging of single object

dragoncoder047:
the new parser I added in #472 is so strict that if you do `debug.log("[boom]")` it crashes the game.

this fixes that, and also causes it to display a dump of the object when you log an object by itself instead of just doing `[object Object]`.


amyspark-ng:
oh that's awesome....

lajbel (approved):

PR #477, opened 2024-10-23
docs: Added many examples for on events

tristanperrow:
Added examples for the following events:
- `onAdd()`
- `onDestroy()`
- `onLoading()`
- `onError()`
- `onResize()`
- `onCleanup()`
- `onGamepadConnect()`
- `onGamepadDisconnect()`
- `onKeyRelease()`
- `onMouseDown()`
- `onMousePress()`
- `onMouseRelease()`
- `onMouseMove()`
- `onHide()`
- `onShow()`
- `onGamepadButtonDown()`
- `onGamepadButtonPress()`
- `onGamepadButtonRelease()`
- `onGamepadStick()`

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -29,6 +29,7 @@ import type {
     BuoyancyEffectorComp,
     BuoyancyEffectorCompOpt,
     CircleComp,
+    CircleCompOpt,
     ColorComp,
     ConstantForceComp,
     ConstantForceCompOpt,
@@ -369,7 +370,18 @@ export interface KAPLAYCtx<
      *
      * @param a The angle to rotate by. Defaults to 0.
      *
-     * @group Components
+     * @example
+     * ```js
+     * let bean = add([
+     *     sprite("bean"),
+     *     rotate(),
+     * ])
+     *
+     * // bean will be upside down!
+     * bean.angle = 180
+     * ```
+
+    * @group Components
      */
     rotate(a?: number): RotateComp;
     /**
@@ -393,6 +405,20 @@ export interface KAPLAYCtx<
     color(): ColorComp;
     /**
      * Sets the opacity of a Game Object (0.0 - 1.0).
+     *
+     * @example
+     * ```js
+     * const bean = add([
+     *     sprite("bean"),
+     *     opacity(0.5) // Make bean 50% transparent
+     * ])
+     *
+     * // Make bean invisible
+     * bean.opacity = 0
+     *
+     * // Make bean fully visible
+     * bean.opacity = 1
+     * ```
      *
      * @group Components
      */
@@ -511,7 +537,7 @@ export interface KAPLAYCtx<
      *
      * @group Components
      */
-    circle(radius: number): CircleComp;
+    circle(radius: number, opt?: CircleCompOpt): CircleComp;
     /**
      * Attach and render an ellipse to a Game Object.
      *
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -1290,18 +1290,82 @@ export interface KAPLAYCtx<
      */
     onDraw(action: () => void): KEventController;
     /**
+     * Register an event that runs when an object with the provided tag is added.
+     * 
+     * @example
+     * ```js
+     * // This will run when the object is added.
+     * onAdd("addTag", () => {
```
lajbel, line 1298: addTag looks as bad example, maybe `bomb`, `enemy` or any real world name

src/types.ts
```diff
@@ -1290,18 +1290,82 @@ export interface KAPLAYCtx<
...
      * @group Events
      */
     onAdd(action: (obj: GameObj) => void): KEventController;
     /**
+     * Register an event that runs when an object with the provided tag is destroyed.
+     * 
+     * @example
+     * ```js
+     * // This will run when the object is destroyed.
+     * onDestroy("destroyTag", () => {
```
lajbel, line 1335: The same as onAdd

src/types.ts
```diff
@@ -1290,18 +1290,82 @@ export interface KAPLAYCtx<
...
+     * Register an event that runs when an object is destroyed.
+     * 
+     * @example
+     * ```js
+     * // This will run when the object is destroyed.
+     * onDestroy(() => {
+     *     debug.log("ohbye")
+     * })
+     * 
+     * let objectToDestroy = add([
```
lajbel, line 1361: the same as onAdd, should be generic names like enemy or bullet

src/types.ts
```diff
@@ -1586,31 +1777,139 @@ export interface KAPLAYCtx<
...
     /**
      * Register an event that runs when user clicks mouse.
+     * 
+     * @example
+     * ```js
+     * // gives cookies on left press, remove on right press
+     * let cookies = 0
+     * onMousePress([`left`, `right`], (m) => {
+     *     if (m == `left`)
+     *         cookies++
```
lajbel, line 1807: please use our format with brackets or change to inline `if(m == "left") cookies++;`

src/types.ts
```diff
@@ -1726,6 +2183,26 @@ export interface KAPLAYCtx<
...
+     * @example
+     * ```js
+     * // player move
+     * let player = add([
+     *     pos(center()),
+     *     anchor(`center`),
+     *     sprite(`bean`),
+     *     area(),
+     *     body(),
+     *     {
```
lajbel, line 2196: Example doesn't need all this components, with pos and sprite is already understable

src/types.ts
```diff
@@ -1717,6 +2140,40 @@ export interface KAPLAYCtx<
     ): KEventController;
     /**
      * Register an event that runs when user releases any gamepad button.
+     *
```
lajbel, line 2143: Complicated, but looks good

src/types.ts
```diff
@@ -1699,6 +2084,29 @@ export interface KAPLAYCtx<
     /**
      * Register an event that runs when user presses any gamepad button.
      *
+     * @example
+     * ```js
+     * // gamepad player controls
+     * setGravity(200)
+     * 
+     * let player = add([
```
lajbel, line 2092: The same, not that much comps needed

src/types.ts
```diff
@@ -1688,6 +2055,24 @@ export interface KAPLAYCtx<
     ): KEventController;
     /**
      * Register an event that runs when user presses certain gamepad button.
+     * 
+     * @example
+     * ```js
+     * // gamepad player jump
+     * setGravity(200)
+     * 
+     * let player = add([
```
lajbel, line 2064: Same about many comps

src/types.ts
```diff
@@ -1679,7 +2024,29 @@ export interface KAPLAYCtx<
     ): KEventController;
     /**
      * Register an event that runs every frame when any gamepad buttons are held down.
-     *
+     * 
+     * @example
+     * ```js
+     * // basic car with brakes
+     * let car = add([
```
lajbel, line 2031: Same about comps, I think we should mantain simple the ones in controller, I like these examples but can be  real examples, not the example of the function, should be simpler

src/types.ts
```diff
@@ -1655,20 +1954,66 @@ export interface KAPLAYCtx<
...
+     * ```
      *
      * @since v3001.0
      * @group Events
      */
     onShow(action: () => void): KEventController;
     /**
      * Register an event that runs every frame when certain gamepad buttons are held down.
+     * 
+     * @example
```
lajbel, line 2000: The same about complexity

lajbel (changes requested):
Many are fine, let some comments

PR #483, opened 2024-10-29
fix: randi now returns 0 or 1 instead of 0 when no args [cherry-pick]

imaginarny:
Cherry picking https://github.com/kaplayjs/kaplay/pull/482 onto this branch


lajbel:
Thanks

lajbel (approved):

PR #506, opened 2024-11-11
update on() example which was incorrect

dragoncoder047:


mflerackers:
The problem is that the callback is attached using the global on, but the trigger is called on the local npc object. Since there is no global trigger function, the on should have been called on the npc.

dragoncoder047:
> the on should have been called on the npc.

No, it works this way too, and you don't need a reference to the npc.

Full example:

```js
kaplay();
loadBean();

const npc = add([
    sprite("bean"),
    pos(100, 100),
    "npc"
]);

on("talk", "npc", (npc, message) => {
    npc.add([
        text(message),
        pos(0, -50),
        lifespan(2),
        opacity(),
    ])
});

onKeyPress("space", () => {
    npc.trigger("talk", "Hello, KAPLAY!");
});

onUpdate(() => {
    npc.pos = vec2(100, 100).add(Vec2.fromAngle(time() * 100).scale(50));
})
```

lajbel (approved):

PR #511, opened 2024-11-14
feat: add better error message to say which components conflict

dragoncoder047:
Screenshot 2024-11-14 at 4 58 29 PM

previously, it would just say "duplicate property width" and you'd be left wondering which components conflict


lajbel (approved):
LGTM

PR #513, opened 2024-11-15
fix: fast loop using update

dragoncoder047:
mostly from https://github.com/kaplayjs/kaplay/issues/35#issuecomment-2126209484 but can run multiple loops per frame, more up to date stuff etc

mostly if the old loop() was had too low an interval, or the game lagged, then the loop() would deadlock on something and freeze the whole browser window, it doesn't seem to do that anymore even when running 10,000 loops per second (>>framerate)

### DIFF
--- a/examples/fastLoop.js
+++ b/examples/fastLoop.js
@@ -1,0 +1,44 @@
+// @ts-check
+
+kaplay();
+
+loadBean();
+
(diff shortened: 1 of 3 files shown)


mflerackers (approved):

lajbel (approved):
LGTM

PR #516, opened 2024-11-16
feat: Adds seek and paused to animate

mflerackers:


src/components/misc/animate.ts
```diff
@@ -550,6 +552,7 @@ export function animate(gopts: AnimateCompOpt = {}): AnimateComp {
             t += dt();
             for (const c of channels) {
                 localFinished = c.update(this as unknown as GameObj<any>, t);
+                console.log("localFinished", localFinished, c.isFinished);
```
lajbel: console.log here?

PR #520, opened 2024-11-18
fix: make health use getters and setters

dragoncoder047:
use modern getters and setters for hp and maxHP


niceEli (approved):

niceEli:
This could be breaking though?

dragoncoder047:
That's why it's probably v4000 only

lajbel:
I have to think this more

lajbel:
Will redo later

PR #535, opened 2024-11-23
"better" docs on map and mapc functions

amyspark-ng:
hopefully this is more understandable

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -2713,15 +2713,19 @@ export interface KAPLAYCtx<
      * @param l1 The minimum value of the source range.
      * @param h1 The minimum result value.
      * @param l2 The maximum value of the source range.
-     * @param h2 The maximum result value
+     * @param h2 The maximum result value.
+     * 
      * @example
      * ```js
      * onUpdate(() => {
-     *      // This variable will be 0 when the mouse is at the left edge and 255 when the mouse is at the right edge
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -2704,7 +2704,24 @@ export interface KAPLAYCtx<
...
+     * @param l2 The maximum value of the source range.
+     * @param h2 The maximum result value
+     * @example
+     * ```js
+     * onUpdate(() => {
+     *      // This variable will be 0 when the mouse is at the left edge and 255 when the mouse is at the right edge
+     *      const redness = map(mousePos().x, 0, width(), 0, 255)
+     *      setBackground(rgb(redness, 0, 0))
+     * })
+     * ```
```
lajbel, line 2725: Space between example and group, also add here `@since` v2000.0 and `@returns`

src/types.ts
```diff
@@ -2704,7 +2704,24 @@ export interface KAPLAYCtx<
...
+     * If the value overshoots, the source range, the result values will also do.
+     * 
+     * For clamping check {@link mapc}
+     * 
+     * @param v The value the function will depend on.
+     * @param l1 The minimum value of the source range.
+     * @param h1 The minimum result value.
+     * @param l2 The maximum value of the source range.
+     * @param h2 The maximum result value
+     * @example
```
lajbel, line 2718: space between param serie and example

lajbel (changes requested):

PR #537, opened 2024-11-24
fix: make offscreen() use global onUpdate(), when pause: true, so it can detect when it comes back onscreen

### LINKED ISSUE #536: bug:`offscreen({pause: true})` gets stuck once paused, must be unpaused manually
```js
kaplay();

loadBean();

const bean = add([
    sprite("bean"),
    pos(),
    offscreen({pause: true}),
    rotate(),
    anchor("center"),
    {
        update() {
            this.angle += 400 * dt();
        }
    }
]);

camPos(0, 0);
wait(1, () => {
    camPos(1000, 1000);
    wait(1, () => {
        camPos(0, 0);
    });
});

onUpdate(() => {
    debug.log(bean.isOffScreen());
});
```

So, the bean is not paused when it is spinning, but as soon as it goes offscreen it pauses cause of the offscreen() comp, but when it comes back on screen again, isOffScreen() returns the right value but bean does not spin cause it is paused. culprit I think is that the offscreen comp uses its own component update() to listen for coming back onscreen which just isn't called when the obj is paused! It should use a global onUpdate?

dragoncoder047:
closes #536

I also added this behavior behind a new flag, unpause, so users that are already using pause: true, won't get broken suddenly


lajbel:
Could you update to latest commits and move the example to playtests?

dragoncoder047:
Bit busy right now but don't merge this pr until I fix the examples and merge master

dragoncoder047:
ok done

Issue #503, opened 2024-11-09
bug: level spatialMap ignores changes to component tilePos


mgmarlow:
## **Describe the bug**

Moving components in a level via `component.moveRight()` (or any directional move function from the `tile` component) does not update a level's internal `spatialMap`, meaning calls to `level.getAt(x, y)` are inaccurate.

The only change detection for spatialMap lives here: https://github.com/kaplayjs/kaplay/blob/master/src/game/level.ts#L122. It works fundamentally based on position, ignoring any changes to `tilePos`.

### What version are you running?

3001/4000.

## **Expected behavior**

Calls to `moveRight()` or other directional move functions from a `tile` component should update the parent level's `spatialMap`, so that `level.getAt(x, y)` returns expected results.

PR #539, opened 2024-11-24
fix: Update level spatial map when tiles are moved

mgmarlow:
## Description

- Expose removeFromSpatialMap and insertIntoSpatialMap from level so they can be called from the tile component.
- When calling a move function from tile, remove the current component from the spatial map and add them back with the new position.
- Contribute a Sokoban example that showcases using tile movement to create a tile-based puzzle game.

Here's what the example looks like:

image

## Related issues

- Closes https://github.com/kaplayjs/kaplay/issues/503

### DIFF
--- a/examples/sokoban.js
+++ b/examples/sokoban.js
@@ -1,0 +1,72 @@
+kaplay({
+    background: [45, 33, 51],
+});
+
+loadSprite("bean", "/sprites/bean.png");
+loadSprite("grass", "/sprites/grass.png");
+loadSprite("steel", "/sprites/steel.png");
+
+const level = addLevel([
+    ".......",
+    ".p   d.",
+    ". b b .",
+    ".     .",
+    ".......",
+], {
+    tileWidth: 64,
+    tileHeight: 64,
+    tiles: {
+        p: () => [sprite("bean"), "player"],
+        b: () => [sprite("grass"), "box"],
+        ".": () => [sprite("steel"), "wall"],
+    },
+});
+
+const player = level.get("player")[0];
+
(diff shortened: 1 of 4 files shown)


mflerackers:
While this fixes one particular use case, it breaks others (for example navigation) and is very costly when the map is large. What should happen is

* level.removeFromSpatialMap(this); // This removes the object from the spatial map at the old tilePos
* update tilePos
* level.insertIntoSpatialMap(this); // This places the object at the correct position in the spatial map at the new tilePos
* level.trigger("spatialMapChanged"); // This makes sure navigation and games that make use of it get notified that the spatial map has changed

mgmarlow:
> While this fixes one particular use case, it breaks others (for example navigation) and is very costly when the map is large.

Ah, fair enough. I wasn't sure whether you preferred exposing both add/remove from the level object as opposed to an invalidate, similar to the navigation API. But I follow the reasoning. Will update it accordingly.

mflerackers:
Best would be to have tile a friend of level, but JavaScript/TypeScript doesn't have any granularity in what's private/public.

mgmarlow:
One thing that's a little ugly in this case is the type cast from `TileComp` to `GameObj` to satisfy the arguments for spatialMap insert/remove:

```ts
level.removeFromSpatialMap(this as unknown as GameObj<TileComp>);
```

I saw the sprite component [doing similar](https://github.com/kaplayjs/kaplay/blob/master/src/components/draw/sprite.ts#L245) so I assumed this is the way to go.

mflerackers (approved):

PR #545, opened 2024-11-28
feat: add k.cancel()

neverUsedGithub:
## Description
Adds support for `k.cancel()`, a new way to cancel events in KAPLAY. As discussed on Discord, `k.cancel` is a method (when it could just be a property) to signify that it has an "action", to maybe help beginners better understand it.

```js
k.onKeyPress((key) => {
    if (key === "q") return k.cancel();
});
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -5510,6 +5510,19 @@ export interface KAPLAYCtx<
      */
     KEventController: typeof KEventController;
     /**
+     * Cancels the event if returned.
+     * 
+     * @since v3001.1
+     * @group Events
+     * @example
+     * ```js
+     * onKeyPress((key) => {
+     *     if (key === "q") return k.cancel();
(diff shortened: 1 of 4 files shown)


niceEli (approved):

src/types.ts
```diff
@@ -5509,6 +5509,19 @@ export interface KAPLAYCtx<
...
+     * @since v3001.1
+     * @group Events
+     * @example
+     * ```js
+     * onKeyPress((key) => {
+     *     if (key === "q") return k.cancel();
+     * });
+     * ```
+     */
+    cancel: () => Symbol,
```
lajbel: Use the correct format described in Contributing.md

lajbel (changes requested):

lajbel (approved):

PR #579, opened 2025-01-03
feat: strict .use()

lajbel:
This changes applies for both versions, v3001 and v4000 and makes strict() only acceept type object or string, otherwise will throw an error. I hope it will help with cases like passing smt is returning undefined, or an extra comma.


mflerackers (commented):
null is also typeof "object"

mflerackers (approved):

PR #580, opened 2025-01-07
chore: Swap To MIT With No-AI Clause(s)

niceEli:
## Description

https://gitingest.com/kaplayjs/kaplay exists and i think we should get some rights against AI companies if they use ingest services to scrape our hard work

Adds 3 clauses to the licence
- You cant use KAPLAY for AI training or else you violate the licence
- KAPLAY is not allowed to be data mined for any AI or neural network purposes, trained on using AI, or used with any neural network
- Anyone who violates clauses 1 and 2 are subject to legal action (you can sue, and you would win since your in the right)

Based on https://github.com/non-ai-licenses/non-ai-licenses/blob/main/NON-AI-MIT with some fixes (like grammar)


mflerackers (approved):

dragoncoder047:
Only problem with this now is that GitHub doesn't autodetect the license, it just says "See license" or something like that instead of "MIT license", so people searching for MIT-licensed game engines will no longer be able to find KAPLAY

PR #581, opened 2025-01-09
feat: added option for sprites to be imported as a separate texture

chqs-git:
Changed the `LoadSpriteOpt` interface to include the `singular` boolean. If this values is equal to true the sprite will be loaded as an separate texture, instead of being placed inside a texture sheet where multiple sprites are placed for optimization on texture loads.

These changes should in no way change the default behaviour.

Preview of changes:
```ts
loadSprite(
  "player",
  "sprites/player.png",
  {
    singular: true
  }
);
```

This pull request is for the v4000 version of kaplay because its closer to a feature rather than a fix. If you think this should be included in the v3001 version please comment.

### DIFF
--- a/src/assets/sprite.ts
+++ b/src/assets/sprite.ts
@@ -67,6 +67,10 @@ export interface LoadSpriteOpt {
      * Animation configuration.
      */
     anims?: SpriteAnims;
+    /**
+     * If the sprite is a single image.
+     */
+    singular?: boolean;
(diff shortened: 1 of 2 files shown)


lajbel:
could you fix the conflicts?, ty!

mflerackers (approved):

lajbel (approved):

PR #592, opened 2025-01-20
feat: css colors

lajbel:
Features:

- `rgb` and `color` now support passing CSS color names, like red or slateblue
- New loader `loadHappy();`


mflerackers (approved):

dragoncoder047:
Just putting all of the colors in a table was the simplest solution and I'm not sure why I didn't just do that -- do we really need to support weird stuff like `color-mix(in hsl, hsl(200 50 80), coral 80%)` tbh.

PR #606, opened 2025-02-12
fix: change collisionIgnore to set to speed up collsion testing

dragoncoder047:
The new playtest is supposed to test the performance of collsionIgnore checking, but for some reason I haven't noticed very much speedup using that test. Maybe it is just a lousy test.

### DIFF
--- a/src/components/physics/area.ts
+++ b/src/components/physics/area.ts
@@ -16,9 +16,8 @@ import {
     Polygon,
     Rect,
     rgb,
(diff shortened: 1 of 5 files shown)


dragoncoder047:
Testing it with my game as a torture test it actually made it slower. Not sure why...

lajbel:
I think this may be outdated and not needed anymore. Reopen if that's not true.

PR #608, opened 2025-02-15
add fixed dt to global options to change it.

dragoncoder047:
It makes it a global option, but maybe it is best to leave it undocumented?


dragoncoder047:
Superseded by #957.

PR #611, opened 2025-02-17
fix: make offscreen component able to change the offscreen distance

dragoncoder047:
if the object size is changed to be greater than the initially specified offscreen distance, then this allows the offscreen distance to be changed so it won't actually disappear until it goes offscreen. additionally the object will use whether its bounding rectangle is truly offscreen if the distance is not specified.

### DIFF
--- a/src/components/transform/offscreen.ts
+++ b/src/components/transform/offscreen.ts
@@ -4,6 +4,7 @@ import { height, width } from "../../gfx";
 import { Rect, testRectPoint, vec2 } from "../../math/math";
 import type { Comp, GameObj } from "../../types";
 import type { KEventController } from "../../utils/";
+import type { RectComp } from "../draw";
 import type { PosComp } from "./pos";
 
 /**
@@ -14,8 +15,11 @@ import type { PosComp } from "./pos";
 export interface OffScreenComp extends Comp {
     /**
      * The minimum distance that the object must be off the screen by to be considered "offscreen".
+     * 
+     * If it is undefined, it means that the object will be considered to be offscreen when its bounding rectangle
+     * (defined by width and height) is not intersecting with the screen rectangle.
      */
-    offscreenDistance: number;
+    offscreenDistance: number | undefined;
     /**
      * If object is currently out of view.
      */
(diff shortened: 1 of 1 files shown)


src/components/transform/offscreen.ts
```diff
@@ -83,15 +90,23 @@ export function offscreen(opt: OffScreenCompOpt = {}): OffScreenComp {
...
             if (!pos) return false;
 
             const screenRect = new Rect(vec2(0), width(), height());
+            if (!this.offscreenDistance && this.width && this.height) {
+                const selfRect = new Rect(pos, this.width, this.height);
+                return selfRect.collides(screenRect);
+            }
+            const sdist = Math.pow(this.offscreenDistance
+                ? this.offscreenDistance
+                : DEF_OFFSCREEN_DIS, 2);
```
lajbel: This is not more complexity added?

dragoncoder047: It only runs for when the offscreen distance is not set and it's a sprite() or rect() or text(), but considering that it allocates a new rect on each frame, I could probably fix that.

lajbel: Yeah, would be cool. It also adds a way for it working nice by default but I'm worried about that allocation.

mflerackers: Do we really need to use pow? it's just 
```ts
const sdist = this.offscreenDistance ? 
    this.offscreenDistance * this.offscreenDistance : 
    DEF_OFFSCREEN_DIS * DEF_OFFSCREEN_DIS;
```

dragoncoder047: But is that faster? If it isn't, I like the shorter solution

mflerackers (approved):

src/components/transform/offscreen.ts
mflerackers: Math.pow needs a function call, and it is written for real numbers, like 20.5^0.32. So I'd does a bit more work than x*x.

dragoncoder047: Is `** 2` the same as pow() or is it faster?

mflerackers: "Math.pow() is equivalent to the [**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Exponentiation) operator, except Math.pow() only accepts numbers."

dragoncoder047: I just did some quick performance benchmarking, and `Math.pow(x, 2)`, `x ** 2`, and `x * x` (where x is a variable) were all the EXACT SAME SPEED.

dragoncoder047: In fact, Math.pow() seems to be marginally faster in the 3 browsers I tested (Chrome, Firefox, and Safari).

https://jsben.ch/Jl8iV

mflerackers: Screenshot 2025-03-04 at 8 56 42

dragoncoder047: Oof, your browser...

This is what I get (and In Firefox, nonetheless...):

Screenshot 2025-03-03 at 6 58 21 PM

mflerackers: Firefox on Mac OS X here.

dragoncoder047: Guess it must be the processor then. Mine's an M3.

mflerackers: Same on chrome on Mac OS X. No idea why a function call + various calculations gets the same speed as a multiplication on you PC.

PR #634, opened 2025-03-12
fix: inspect view now shows the tags even after the tagsAsComponents refactor

dragoncoder047:
They now show as `tag: happy` instead of just `happy`, but I think that's more clear that it's a tag and not a component, now that tags are not emulated with anonymous components.

I think it's relatively unlikely that someone is going to create a custom component and give it the id "tag" and also have it make an identical inspect() value to what tags show as here.


lajbel (approved):
LGTM ❤️

PR #635, opened 2025-03-12
fix: only select an object for inspecting if the object is not paused

dragoncoder047:
I have a whole bunch of levels that are swapped in and out by pausing and unpausing them, but the debug inspect doesn't look at whether the object is paused, and the areas from the first level basically "cover" the active level and prevent me from inspecting it.

This changes it so it only inspects the unpaused objects.

@lajbel should this be changeable by a global option?


lajbel:
I think it should be optionable

dragoncoder047:
> I think it should be optionable

Done!

lajbel (approved):

PR #662, opened 2025-03-24
feat: tuple component list

### LINKED ISSUE #645: bug: empty components can break an object's type
## **Describe the bug**
When you add an object (or a component) with no properties in an object it will break its type, sometimes listing it as simply GameObject or any (not confirmed)

This also happens when a property is any i think (not confirmed either)
```ts
// breaks type info for game objects
export const myComponent = () => {
  return {
    id: 'foo',
    add() {},
    // Uncommenting the property fixes type info
    // foo: true,
  };
};
```

lajbel:
## Description

This makes `add()` and `make()` convert their arrays to tuples so types like { a: string, b: string } | { a: string } are not converted to { a: string } but mantained as an Union. Also it add comments to some of the weirdest KAPLAY TypeScript methods.

I know I added much `as` in the make code but I don't think really worth making it correctly, only will slow types. This is only for user end

### Issues or related

- Closes #645

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   (`GJK`) distance algorithm.
 - Changed default behaviour of `kaplay({ tagsAsComponents: false })` to `false`.
 - Now if you pass a nullish value to `.use()` it throws an error
+- Improved TypeScript in game objects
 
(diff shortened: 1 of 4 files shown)


mflerackers (approved):

dragoncoder047:
noticed that the `const` modifier is deep, so if you include an object `{eaten: false}` inline, it winds up with the type being `{readonly eaten: boolean}`, even though it's not, how to fix?

PR #675, opened 2025-03-26
refactor: Part 1

lajbel:
## Description

Okay I know how this look, let me explain.

I think for add these features #674 and #673, we need a better organized and documented code. Know what depends of what and have the ability of separate all in concerns and features for adding tree shaking in the future. So here's my idea:

- Organize folders
- Document
- Making tests

### What this PR does?

This basically takes all of the `initXXXX` and moves them to `core/engine.ts` and `createEngine()`. Maybe in the future we can merge all the states in 1. 

Also the collision system is moved to `ecs/systems/collision.ts` also for organization purposes. I think we can use the `ecs` folder for moving `make.ts` and `components` in the future.

Some files and functions (as I commented in the code) should be in other folders, for example we can have a `rendering/` folder for all rendering in general. Well, this is the start.

### DIFF
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -1,9 +1,10 @@
-// everything related to canvas, game loop and input
+// App is everything related to canvas, game loop and input
 
 import type {
     Cursor,
     GamepadDef,
     GamepadStick,
+    KAPLAYOpt,
     Key,
     KGamepad,
     KGamepadButton,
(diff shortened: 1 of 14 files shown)


lajbel:
I tried to edit the less quantity of files anyway. But I also modified some `init` functions for receive opt directly instead of their bunch of options. It makes the `startEngine` code being very legible

mflerackers (approved):

PR #676, opened 2025-03-26
feat: allow changing font mid-text and add stretch in place option

### LINKED ISSUE #620: feat: change font in text component
I already mentioned this on the discord but it would probably be a nice addition.

The text styles entries of the text component can have a `font` property that change the font, so there can be multiple fonts in the same block of text.

To minimize the texture switching the quads for all of the letters can probably just can be sorted by font, and rendered in that order, since the quads don't overlap and won't be translated by a sort.

dragoncoder047:
closes #620

you can change the font mid-text, and also prevent scale from making the text box wider.


lajbel:
Awesome

PR #696, opened 2025-04-06
fix: some examples were bugged

dragoncoder047:
* the loop example explicitly passed in a loop count of -1, which is not supported anymore to mean "infinity".
* the raycast examples still used unuse when they meant untag, which resulted in the onMouseRelease handler not releasing the selected objects and they got stuck to the mouse.


lajbel (approved):

lajbel:
Awesome, thank you so much!

PR #710, opened 2025-04-09
fix: dialog example was broken

dragoncoder047:
change renderedText to formatted text property


mflerackers (approved):

lajbel:
I fixed on described examples

Issue #724, opened 2025-04-12
bug: CapsLock doesn't work with `textInput()`


amyspark-ng:
### Describe the bug.

i should make this field optional

### Summary

- [ ] Fixed in Master
- [ ] Fixed in v4000
- [ ] Fixed in v3001

dragoncoder047:
I guess all we need to do is look at the caps lock state and upper it if it is on

```js
document.body.onkeydown = e => {
  console.log(e.getModifierState("CapsLock") ? "caps lock is on" : "caps lock is off")
}
```

maybe it would also be cool to have the onKeyPress handler expose the modifier states - caps lock , alt, control, shift, etc. to make keyboard shortcuts possible

amyspark-ng:
this is pretty useful! is was making the issue to remind myself to fix it later

well, you can always do something like this i think?

```ts
onKeyPress("s", () => {
    if (isKeyDown("control")) //do action
})```

**Comment by dragoncoder047** [comment]
only problem is we need to possibly add a way for prevent default on the event if needed

```js
onKeyPress("s", () => {
    if (isKeyDown("meta")) debug.log("hi")
})
```

you press command+s (on mac) and you get a hi and a browser save dialog

amyspark-ng:
in some games where i need to prevent defaults like "ctrl + s" i do something like this
```ts
// prevent ctrl + s weirdness
document.addEventListener("keydown", function(e) {
	if (e.key === "s" && (navigator.userAgent.includes("Mac") ? e.metaKey : e.ctrlKey)) {
		e.preventDefault();
	}
}, false);
```

i don't know if adding a kaplay-way of preventing defaults is necessary, since is not something really related to kaplay itself

dragoncoder047:
could also do something where you declare your key binds as a button and then kaplay preventDefault's them automatically, but only if they are defined as a button

```js
kaplay({
    buttons: { save: { keyboard: ["meta+s"] } },
})
```

amyspark-ng:
sounds good, kinda like a niche feature, but good

PR #726, opened 2025-04-12
feat: mayus

amyspark-ng:
closes https://github.com/kaplayjs/kaplay/issues/724

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,7 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   non-Promise-based API return value - @imaginarny
 - changing `debug.timeScale` now actually makes the game change speed by
   affecting `dt()` - @lajbel
+- CapsLock now affects textInput() - @amyspark-ng
 
 ### Removed
 
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -110,6 +110,7 @@ export const initAppState = (opt: {
         skipTime: false,
         isHidden: false,
         numFrames: 0,
+        capsOn: false,
         mousePos: new Vec2(0),
         mouseDeltaPos: new Vec2(0),
         keyState: new ButtonState<Key>(),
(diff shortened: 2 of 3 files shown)


amyspark-ng:
i just realized this creates an event listener onupdate!! sorry im foggy, where should i put this document.body.onkeydown()?

dragoncoder047:
you know this doesn't have to be document.body.onkeydown, that was just a quick and dirty example. probably attaching to state.canvas.addEventListener("keydown"), and I already think there is a handler for that somewhere else that it could be put in

amyspark-ng:
yeah there was a better way looking at the code, hopefully this is fine

lajbel:
Changelog it for v4000

dragoncoder047:
wondering what the behavior should be if shift is held while caps lock is on, should it unshift `aND tYPE lIKE tHIS`?

amyspark-ng:
i think it's built in that they counteract, so i can't do much about it i think??

dragoncoder047:
your code uppers it if either capsLock or shift is down (normal OR), not if *only* one is down (exclusive OR)

amyspark-ng:
ah you're right lemme fix

amyspark-ng:
if (
                        (_k.k.isKeyDown("shift") && !_k.app.state.capsOn)
                        || (!_k.k.isKeyDown("shift") && _k.app.state.capsOn)
                    ) {
                        this.typedText += character.toUpperCase();
                    }
                    
                    is this fine?

dragoncoder047:
that technically works, but you can probably use *the XOR operator* `^` cause true acts like 1, false acts like 0, and 1 is truthy and 0 is falsy

amyspark-ng:
ah this Boolean(!a ^ !b)?

dragoncoder047:
no just ((a) ^ (b)) it casts automatically

and anyway,

amyspark-ng:
well that causes a type error
"The ^ operator is not allowed for boolean types. Consider using !== instead."
Doing boolean() also causes an error

dragoncoder047:
woops I also forgot that if the operands are boolean then !== is the same as ^

amyspark-ng:
checked and it works now properly, thank you very much for the help

mflerackers (approved):

PR #728, opened 2025-04-12
refactor: (Described Examples Part 2) remove assets, move to examples/, simpler path

lajbel:
MF probably is asking when he merged part 1

In summary:
- Removed `assets` from the root, so all it's now inside `/examples`
- Now for import stuff inside /examples/sprites you only have to write /sprites, so path is simpler rn


lajbel:
Test fails because they are dumb

lajbel:
Will fix tomorrow, also the reason for removing assets folder is because there's already crew

lajbel:
Like if you're reading this on Discord

lajbel:
I addressed all FIXME cases

mflerackers (approved):

PR #730, opened 2025-04-13
Add example of defining frames by quad input.

nojaf:

I was a bit confused yesterday about defining frames of a sprite by direct input using `Quad`.
Wanted to add an example for further users.

## Summary

- [ ] Changeloged

Let me know if this needs a changelog.

### DIFF
--- a/examples/sprite_frames.js
+++ b/examples/sprite_frames.js
@@ -1,0 +1,92 @@
+/**
+ * @file Sprite Frames
+ * @description How to load individual sprite frames in KAPLAY.
+ * @difficulty 0
+ * @tags basics, animation
+ * @minver 3001.0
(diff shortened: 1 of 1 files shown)


amyspark-ng:
i don't think changelog for examples is necessary

nojaf:
Another thing that could have helped me is an ArgumentException in 
https://github.com/kaplayjs/kaplay/blob/b5f7cfde0af2bb1b648e30c7d91e3217d88db6a1/src/math/math.ts#L100-L105
if any of these values were not between 0 and 1.
Not sure if that comes with a performance cost though.

lajbel:
Will close as I think we have to discuss it a little more. Will discuss it internally

PR #735, opened 2025-04-18
fix: PatrolComp is not going to last waypoint

### LINKED ISSUE #734: bug: PatrolComp is not going to last waypoint
### Bug Description

Consider this small example:

```js
const kat =
  add([
      sprite("kat"),
      pos(40, 30),
      patrol({
          waypoints:[
              vec2(100,100),
              vec2(120, 170),
              vec2(50,50),
              // never reached
              vec2(1000,1000)
          ]
      }),
      "kat"
  ]);

kat.onPatrolFinished(kat => {
    // kat is undefined
    debug.log(kat)
})
```

I think the problem is in:
https://github.com/kaplayjs/kaplay/blob/eac9fae437ecc62128c2d40fc89e5ae25a8b6986/src/ecs/components/level/patrol.ts#L94-L95

The `Math.min` is already checking if it is not the last index or not.

Also, the `onPatrolFinished` callback types are wrong.
https://github.com/kaplayjs/kaplay/blob/eac9fae437ecc62128c2d40fc89e5ae25a8b6986/src/ecs/components/level/patrol.ts#L97

Is not passing down any arguments, so the `gameObj` is undefined.

### Version

master

### Playground Link

https://play.kaplayjs.com/?code=eJxtUFELwiAYfPdXSE8KsrnVCIp67Ln3sQebtqShQ90iYv89tTUodiB8353ecdZaWQfvzMEDYJyjEkAP2xnpBFp5foVJpDpt0YYSuKZfgjmjW%2FSKS8CDPTstlbO7cuYCBlHnKKOU%2BDO9%2FZVyb5ttF7WCkuJfSFOoxCAMNILVN8EXw2IaxbNWxWmcrGIxUOE9AH5KtDrHMieppPWOKP7HEX66%2BbywSwt7xcVVqimSi0vfJK1uwnUMRvwGDrJPLw%3D%3D&version=master

### Extra information

_No response_

### Summary

- [x] Fixed in v4000
- [ ] Fixed in v3001

nojaf:

Fixes #734

## Summary

- [x] Changeloged

### DIFF
--- a/src/ecs/components/level/patrol.ts
+++ b/src/ecs/components/level/patrol.ts
@@ -91,10 +91,12 @@ export function patrol(
                         }
                         break;
                     case "stop":
-                        index = Math.min(index + 1, waypoints.length - 1);
-                        if (index == waypoints.length - 1) {
+                        if (index < waypoints.length - 1) {
+                            index += 1;
+                        }
+                        else if (!finished) {
                             finished = true;
-                            this.trigger("patrolFinished");
+                            this.trigger("patrolFinished", this);
                         }
                         break;
                 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,8 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - changing `debug.timeScale` now actually makes the game change speed by
   affecting `dt()` - @lajbel
 - CapsLock now affects textInput() - @amyspark-ng
+- PatrolComp is not going to last waypoint
+  ([#734](https://github.com/kaplayjs/kaplay/issues/734)) - @nojaf
 
 ### Removed
 
--- a/examples/patrol.js
+++ b/examples/patrol.js
@@ -1,0 +1,27 @@
+/**
+ * @file Patrol
+ * @description How to patrol a sprite.
+ * @difficulty 1
+ * @tags physics
+ * @minver 3001.0
+ */
+kaplay();
+
+loadBean();
+
+const bean = add([
+    sprite("bean"),
+    pos(40, 30),
+    patrol({
+        waypoints: [
+            vec2(100, 100),
+            vec2(120, 170),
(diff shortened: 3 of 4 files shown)


mflerackers (approved):

src/ecs/components/level/patrol.ts
```diff
@@ -91,10 +91,12 @@ export function patrol(
...
                     case "stop":
-                        index = Math.min(index + 1, waypoints.length - 1);
-                        if (index == waypoints.length - 1) {
+                        if (index < waypoints.length - 1) {
+                            index += 1;
+                        }
+                        else if (!finished) {
                             finished = true;
-                            this.trigger("patrolFinished");
+                            this.trigger("patrolFinished", this);
```
KeSuave, line 99: isn't it a bit redundant to send `this`?

```ts
gameObj.onPatrolFinished(() => debug.log(`gameObj new pos: ${gameObj.pos.x} ${gameObj,pos.y}`))
```

nojaf: It is matching the current interface. I can also update the interface of course if you prefer.

mflerackers: Not if you attach the same function, not a lambda closure, to each patrolling unit.

KeSuave: there might be a reason for it, mostly curious as it does feel redundant

mflerackers:
```ts
function unitPatrolFinished(unit) {
  // Patrol finished, rest for 50 seconds then restart patrol
}

function createPatrol() {
  const patrol = add([...]);

  patrol.onPatrolFinished(unitPatrolFinished);

  return patrol;
}
```

It is because creating a new anonymous lambda with its own closure for each object is not really performant, so advanced users may want to use a function instead. In that case they want to know which object finished patroling.

PR #736, opened 2025-04-20
Layers, transform update, stack functions, arc points, step and smoothstep

mflerackers:
- Layers now work globally, no longer only between siblings.
- Areas no longer struggle with parents whose transform inst't up-to-date.
- Exported step and smoothstep.
- Small circles and arcs use now less points than larger ones.
- Added pushMatrix, storeMatrix and loadIdentity to the stack functions.


lajbel (approved):

amyspark-ng:
R.I.P #523 , we won't miss you

imaginarny:
> R.I.P #523 , we won't miss you

Although it's still broken 😐

amyspark-ng:
yes hovers are broken but that's a whole sepparate issue 😁

imaginarny:
I mean ye, the issue is separate, but what's shown in the video exactly wasn't fixed xD.. also, we should comment there instead xD

PR #743, opened 2025-04-21
feat: `GameObjRaw` as a prototype, and others

lajbel:

- Add `GameObjRaw` as a prototype, so all objects uses it

## Summary

- [ ] ~Changeloged~ Not neccesary


lajbel:
I've removed the components. Maybe I have to think if it makes sense, if it makes sense probably we have to upgrade component api to add `use()` and `unuse()` as we need this running when the component is used (and not when it's added) for it really being setting state before it being added

amyspark-ng:
add() runs when the object is added and not when the component is used?

lajbel:
This code basically creates an event handler for the 
"add" event of an object, calling the add() from the component: https://github.com/kaplayjs/kaplay/blob/master/src/ecs/make.ts#L463-L474

May not be clear, but the call is here: https://github.com/kaplayjs/kaplay/blob/master/src/ecs/make.ts#L473

mflerackers (approved):

amyspark-ng:
a guide for this or a docs would be pretty useful i think...

PR #758, opened 2025-04-27
examples: fixed some grammar on lerp.js

amyspark-ng:
its-it-is-difference
/jk


lajbel:
Thanks Amy for helping me improve my English

Issue #759, opened 2025-04-27
feat: some health() fixes


amyspark-ng:
### Is your feature request related to a problem? Please describe.

Add HealthComp.dead

also if you run HealthComp.hp() on HealthComp.onDead() it will not return 0, i experienced something like this

PR #765, opened 2025-04-30
Feat: HealthComp changes

amyspark-ng:
fixes https://github.com/kaplayjs/kaplay/issues/759
## Summary
Didn't have time to actually make it work, will finish hopefully in a few hours

Makes me wonder, should we remove hurt() and heal() and only run onHeal and onHurt if the original hp is lower or higher when doing an hp setter?

- [ ] Changeloged

### DIFF
--- a/src/ecs/components/misc/health.ts
+++ b/src/ecs/components/misc/health.ts
@@ -1,6 +1,4 @@
 import type { KEventController } from "../../../events/events";
-import { _k } from "../../../kaplay";
-import { clamp } from "../../../math/clamp";
 import type { Comp, GameObj } from "../../../types";
 
 /**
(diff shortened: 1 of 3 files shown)


src/ecs/components/misc/health.ts
```diff
@@ -18,19 +20,15 @@ export interface HealthComp extends Comp {
...
      * Max amount of HP.
      */
-    maxHP(): number | null;
+    maxHP: number | null;
     /**
-     * Set max amount of HP.
+     * Wheter hp is 0.
      */
-    setMaxHP(hp: number): void;
+    dead: boolean;
```
dragoncoder047: shouldn't this be readonly cause you have no setter

amyspark-ng: true true, i had forgotten you can do that

src/ecs/components/misc/health.ts
```diff
@@ -61,30 +59,24 @@ export function health(
...
+            this.hp = maxHP ? Math.min(maxHP, this.hp - n) : this.hp - n;
             this.trigger("hurt", n);
+            if (this.hp <= 0) {
+                this.trigger("death");
+            }
         },
         heal(this: GameObj, n: number = 1) {
             const origHP = hp;
+            this.hp += n;
             this.setHP(hp + n);
```
dragoncoder047: also, shouldn't this be removed now

amyspark-ng: yes it's wip, im at uni so i didn't have time to finish it when i started i also asked about it in dev to know opinion on removing heal and hurt

src/ecs/components/misc/health.ts
lajbel: Yeah it should be readonly

lajbel (commented):
Good, remember update the example that is failing

PR #772, opened 2025-05-07
fix: hp vs. hp() after #765

dragoncoder047:
the hp was changed to a getter and not a function but the examples weren't updated so examples were crashing.


amyspark-ng:
dragoncoder done fixed my pr call that a team effort

PR #776, opened 2025-05-13
fix: update wasn't making effect at all and pause events even if created after .paused = true

lajbel:
+feat: now you can use /crew urls in dev mode

Like if you love Git Blame


mflerackers (approved):
Though next time try to not add unrelated changes.

PR #779, opened 2025-05-16
fix: use an object event instead of an app event in AreaComp.onClick

lajbel:
Fixed `AreaComp#onClick()` attaching events to app, instead of object, so
  event wasn't being paused with `obj.paused`

## Summary

- [x] Changeloged


lajbel:
what

amyspark-ng:
no te hagas el gringo....

mflerackers (approved):

PR #782, opened 2025-05-17
fix: solve touch positions not being transformed to viewport

lajbel:
This PR fixes the letterbox/scale support with touch functions. 

## Summary

- [x] Changeloged
- [x] Backported to v3001


mflerackers (changes requested):
If there are two fingers tracking, is it correct to always pass only the transformed position of the first finger? I don't think so. Multi-finger gestures will break that way.

lajbel:
So dumb, true

mflerackers (approved):

PR #784, opened 2025-05-18
feat: Cache matrix inverse, transform point in hasPoint instead of transforming area

mflerackers:


dragoncoder047:
since inverse of inverse is the original matrix shouldn't `m.inverse.inverse === m` rather than a copy of `m`? or is it not worth it to do that

mflerackers:
I thought about that, but I couldn't find a real use case for that. And I'd rather not make a circular reference (which I would have two break on both sides when clearing).

lajbel (approved):

PR #799, opened 2025-06-05
fix: textInput Component backspace

KeSuave:
## Summary

Objects with textInput with non-focused get cleared when hitting backspace. Only those with focus should be cleared.

- [x] Changeloged


src/ecs/components/misc/textInput.ts
```diff
@@ -108,8 +108,8 @@ export function textInput(
             backEv = _k.app.onKeyPressRepeat("backspace", () => {
                 if (this.hasFocus) {
                     this.typedText = this.typedText.slice(0, -1);
+                    flip();
```
dragoncoder047, line 111: Does it really get cleared if not focused? flip() does not change the typedText property, which is the master copy of the text input value. All I can see this doing is triggering the "input" event when it's not focused. But good catch.

KeSuave: flip changes text:

```ts
             const flip = () => {
                this.text = this.typedText.replace(/([\[\\])/g, "\\$1");
                this.trigger("input");
            };
```
            
therefore it clears if you have not typed anything, but still have text in the text component

dragoncoder047: isn't that supposed to happen, then? the textInput component controls the text property, so having it cleared when this.typedText is empty is what is supposed to happen. if you want to modify the text yourself you should be removing the textInput component.

KeSuave: it shouldnt happen if it is not focused, if it is then yes

dragoncoder047: I think then a better title for this PR would be then "fix: textInput updating when not focused" because if the thing has `typedText = "foo"` and `text = "bar"` an you hit backspace, it will update it to `text = "foo"`, not clear it. It only clears if typedText is already cleared !

KeSuave: sure, if the team needs a better description of the fix and feel yours or their title would be better i can change it

lajbel:
Can you do a repro PR for v3001? Thank yo sou much for your contribution :D:D

PR #811, opened 2025-06-12
fix: textInput Component backspace v3001

KeSuave:
## Summary

Objects with textInput with non-focused get cleared when hitting backspace. Only those with focus should be cleared.


lajbel:
Thank you

PR #826, opened 2025-07-01
feat: work on prefabs

amyspark-ng:
wip

### DIFF
--- a/src/ecs/components/draw/outline.ts
+++ b/src/ecs/components/draw/outline.ts
@@ -3,20 +3,29 @@ import { Color, rgb } from "../../../math/color";
 import type { Comp, Outline } from "../../../types";
 
 /**
+ * The serialized {@link outline `outline()`} component.
+ *
+ * @group Component Serializations
+ */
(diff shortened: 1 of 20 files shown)


src/ecs/components/draw/outline.ts
```diff
@@ -32,5 +40,19 @@ export function outline(
...
+            join: this.outline.join,
+            miterLimit: this.outline.miterLimit,
+            opacity: this.outline.opacity,
+            width: this.outline.width
+          }  
+        },
     };
 }
+
+export function rectFactory(data: ReturnType<OutlineComp["serialize"]>) {
```
mflerackers: Should be outlineFactory i think

amyspark-ng:
i think it's good enough now? tell me if anything is missing that i could do

i didn't add serializing for some components that i deemed too complex and didn't want to do anything without asking (polygon, drawon, particles, uvquad, general rendering components etc etc)

lajbel:
ty

PR #833, opened 2025-07-04
feat: export more stuff

dragoncoder047:
exports Collision and AnchorPt


lajbel:
`anchorPt` seems strage tho, maybe `anchorToVec2` ?

mflerackers:
anchorPt is the old kaboom name, but since it was never exported, it's fine to rename it.

lajbel (approved):

PR #838, opened 2025-07-05
fix: clear inputEvents in GameObj.clearEvents()

amyspark-ng:
is it a fix?
should it be changelogged?
is it fine?


mflerackers (changes requested):
You cancel the first element, then throw away the last element. Then you cancel the second element, and throw away the last one, etc. I don't think that it what you want.
You either need to iterator from the end to the start, or need to use the result of pop, like while (this._inputEvents.length) { this._inputEvents.pop().cancel(); }

mflerackers (approved):

PR #839, opened 2025-07-06
feat: implemented directions and toAxis for Vec3

skyebleuu:
made UP, DOWN, LEFT, RIGHT, FORWARDS, BACKWARDS, ZERO, and ONE values and also `toAxis()`


dragoncoder047:
lol I don't think Vec3 is used anywhere yet other than being exported

mflerackers (approved):

PR #842, opened 2025-07-09
feat: Push and Pop Scenes methods

itzKiwiSky:

This changes add methods for pushing and popping scenes from a stack system, is good for page behaviour

```ts
pushScene(id: string, args: ...unknown[])
```

```ts
popScene()
```

check `scenepushandpop.js` test for examples

## Summary

pls tell me if need changelog or any change

### DIFF
--- a/src/game/game.ts
+++ b/src/game/game.ts
@@ -48,6 +48,10 @@ export type Game = {
      */
     sceneStack: Array<SceneState>;
     /**
+     * The current active scene arguments
+     */
+    currentSceneArgs: unknown[];
+    /**
      * The current scene of the game.
      */
     currentScene: string | null;
@@ -144,6 +148,7 @@ export const createGame = (): Game => {
             transform: new Mat23(),
         },
 
+        currentSceneArgs: [], // stores the current scene arguments //
         sceneStack: [], // stores the scene names //
 
         // Systems
--- a/src/game/game.ts
+++ b/src/game/game.ts
@@ -44,7 +44,7 @@ export type Game = {
      */
     scenes: Record<SceneName, SceneDef>;
     /**
-     * The scene stack that stores the scene name
+     * The scene stack that stores the scene states
      */
     sceneStack: Array<SceneState>;
     /**
--- a/src/game/game.ts
+++ b/src/game/game.ts
@@ -44,6 +44,10 @@ export type Game = {
      */
     scenes: Record<SceneName, SceneDef>;
     /**
+     * The scene stack that stores the scene name
+     */
+    sceneStack: Array<string | null>;
+    /**
      * The current scene of the game.
      */
     currentScene: string | null;
@@ -139,6 +143,8 @@ export const createGame = (): Game => {
             shake: 0,
             transform: new Mat23(),
         },
+
+        sceneStack: [], // stores the scene names //
 
         // Systems
         systems: [], // all systems added
--- a/src/game/game.ts
+++ b/src/game/game.ts
@@ -17,7 +17,7 @@ import { KEventHandler } from "../events/events";
 import { Mat23, RNG } from "../math/math";
 import { Vec2 } from "../math/Vec2";
 import type { GameObj } from "../types";
-import type { SceneDef, SceneName } from "./scenes";
+import type { SceneDef, SceneName, SceneState } from "./scenes";
 
 /**
  * The "Game" it's all the state related to the game running
@@ -46,7 +46,7 @@ export type Game = {
     /**
      * The scene stack that stores the scene name
      */
-    sceneStack: Array<string | null>;
+    sceneStack: Array<SceneState>;
     /**
      * The current scene of the game.
      */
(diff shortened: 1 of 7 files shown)


src/game/scenes.ts
```diff
@@ -53,6 +57,26 @@ export function go(name: SceneName, ...args: unknown[]) {
...
+    _k.game.sceneStack.push(_k.game.currentScene);
+    go(id, args);
+
+}
+
+export function popScene()
+{
+    const sceneData: any = _k.game.sceneStack.pop();
+    if (sceneData === undefined)
+        return;
```
mflerackers: What is best behavior here? Throw an exception or fail silently?

src/game/scenes.ts
```diff
@@ -8,6 +8,10 @@ import { _k } from "../shared";
  */
 export type SceneName = string;
 export type SceneDef = (...args: any) => void;
+export type SceneState = {
```
mflerackers, line 11: This is currently unused?

src/game/game.ts
```diff
@@ -140,6 +144,8 @@ export const createGame = (): Game => {
             transform: new Mat23(),
         },
 
+        sceneStack: [], // stores the scene names //
```
mflerackers, line 152: I think it should store SceneState items, not strings

src/game/scenes.ts
```diff
@@ -53,6 +57,26 @@ export function go(name: SceneName, ...args: unknown[]) {
     _k.game.currentScene = name;
 }
 
+export function pushScene(id: SceneName, ...args: unknown[])
+{
+    //if (_k.game.scenes[_k.game.currentScene])
+    if (_k.game.scenes[id] === undefined)
+        return;
+    
+    _k.game.sceneStack.push(_k.game.currentScene);
```
mflerackers: The args aren't being pushed, which will make pop fail, since it can't pass the args.

src/game/scenes.ts
```diff
@@ -53,6 +57,26 @@ export function go(name: SceneName, ...args: unknown[]) {
...
+
+}
+
+export function popScene()
+{
+    const sceneData: any = _k.game.sceneStack.pop();
+    if (sceneData === undefined)
+        return;
+
+    go(sceneData);
```
mflerackers: Only the name is passed, not the args.

itzKiwiSky:
I need to mention I added another function as well

now has:

- `pushScene()`
- `popScene()`
- `getSceneArgs()`

src/game/game.ts
```diff
@@ -43,6 +43,14 @@ export type Game = {
      * The scenes of the game.
      */
     scenes: Record<SceneName, SceneDef>;
+    /**
+     * The scene stack that stores the scene name
```
mflerackers: Should be scene state, not name

src/game/scenes.ts
```diff
@@ -47,12 +51,37 @@ export function go(name: SceneName, ...args: unknown[]) {
             transform: new Mat23(),
         };
 
+        // clear the arguments //
+        _k.game.currentSceneArgs.splice(0, _k.game.currentSceneArgs.length);
```
mflerackers: I think you can just set length to 0

src/game/scenes.ts
```diff
@@ -47,12 +51,37 @@ export function go(name: SceneName, ...args: unknown[]) {
...
+        args: _k.game.currentSceneArgs,
+    });
+    go(id, args);
+    return;
+}
+
+export function popScene()
+{
+    const sceneData: SceneState | undefined  = _k.game.sceneStack.pop();
+    console.log(sceneData);
```
mflerackers: This is debugging code

src/game/scenes.ts
```diff
@@ -47,12 +51,37 @@ export function go(name: SceneName, ...args: unknown[]) {
...
+    if (sceneData === undefined)
+        throw new Error("No more scenes to pop!");
+
+    go(sceneData.sceneID ?? "", sceneData.args);
```
mflerackers: When is scene id null? At the start? Because if there is no concept of a default scene, go("") will fail.

mflerackers (changes requested):

mflerackers (approved):

lajbel:
Did you really think it was going to be that easy? Contribute to an Open Source repository? Ha, ha, ha. Now comes the final boss, my approval, and it won't be easy at all...

itzKiwiSky:
> Did you really think it was going to be that easy? Contribute to an Open Source repository? Ha, ha, ha. Now comes the final boss, my approval, and it won't be easy at all...

💀 ☠️

src/game/scenes.ts
```diff
@@ -47,12 +51,40 @@ export function go(name: SceneName, ...args: unknown[]) {
             transform: new Mat23(),
         };
 
+        // clear the arguments //
+        _k.game.currentSceneArgs.length = 0;
```
mflerackers: Actually don't do this. Since it is not a copy of the args, so you might be clearing an array which is on your stack.

lajbel:
Checks are not ran anymore?

mflerackers (approved):

lajbel (changes requested):
Run `pnpm fmt`

lajbel:
Awesome work, thank you so much ❤️✨ 

beaning

lajbel (approved):

PR #846, opened 2025-07-21
feat: add pause, unpause, show, hide events & hook it into sweepandprune

### LINKED ISSUE #844: feat: onPause and onUnpause events

I had another idea tangentially related to physics. The objects that are paused are not physics-simulated, but if they exist they are still being checked by the sweepandprune for whether they are paused or not, so if there are a lot of paused objects (which there are in my game) it might help to detect when objects are paused and unpaused and remove and re-add them to the sweepandprune, so it doesn't even waste time iterating over them.

### Any more information?

I'm not sure exactly how to do this, but either way it should work hierarchically - if you pause/unpause the parent the onPause/onUnpause event fires on both the parent and all of its descendants (unless they're already paused).

...maybe also add onHide and onShow the same way

dragoncoder047:

closes #844 mostly.

* Adds onPause, onUnpause, onHide, onUnhide
* renames conflicting global event hooks (& onResize for good measure too)
* fixes the sweepandprune to  not create and destroy an iterator on every frame (by letting the sweepandprune drive the iteration instead of, y'know, using the *iterator* to basically have 2 loops controlling the iteration unnecessarily)

## Summary

- [X] Changeloged

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,17 +1,6 @@
 {
   "typescript.tsdk": "node_modules/typescript/lib",
   "typescript.preferences.importModuleSpecifierEnding": "minimal",
-  "exportall.config.folderListener": [
-    "/src/app",
-    "/src/assets",
-    "/src/components",
-    "/src/components/draw",
-    "/src/components/level",
-    "/src/components/misc",
-    "/src/components/physics",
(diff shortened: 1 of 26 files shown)


lajbel:
This PR seems outdated. Reopen if it's still being worked.

PR #849, opened 2025-07-24
fix: make exists() work for nested objects

dragoncoder047:
I had some weird bugs crop up because I had a tree of objects, removed the topmost object, and then asked the children objects whether they exist. They aren't attached to the scene graph anymore so they don't interact with anything, but `child.exists()` was still returning true, and this was confusing my code. Now it recursively sets all the levels' parents to null when an object is destroyed.

## Summary

- [x] Changeloged

### DIFF
--- a/src/ecs/entity/GameObjRaw.ts
+++ b/src/ecs/entity/GameObjRaw.ts
@@ -582,6 +582,10 @@ export const GameObjRawPrototype: Omit<InternalGameObjRaw, AppEvents> = {
         // We assume this will never be ran in root
         // so this is GameObj
 
+        if (this.id === null) {
+            throw new Error("Can't re-parent destroyed object");
+        }
+
         if (this._parent === p) return;
         const index = this._parent
             ? this._parent.children.indexOf(this as unknown as GameObj)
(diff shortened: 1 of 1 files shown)


mflerackers:
I think the best way is to remove the id recursively when detaching from the tree. When it is reattached, it should get a new id.

dragoncoder047:
> I think the best way is to remove the id recursively when detaching from the tree. When it is reattached, it should get a new id.

not sure what you meant here - I changed it so that when yo destroy() an object it destroy()s all of its children recursively so they all have parent === null and the original check works.

Do you mean checking if id !== null instead and just recursively set id to null? I am not sure how the whole "reattaching" thing would work because add() doesn't take existing objects anymore, it can only create new.

mflerackers:
I know you can't reattach a destroyed object because well, components will have ran their destroy handlers and they won't work anymore. Though it might be OK in the future to set parent to null to temporarily detach a sub-tree. In that case parent is a bad measure of knowing whether an object is destroyed or not. But id would be. A destroyed object would no longer have an id, while a sub-tree waiting to be reattached would still have its id.

dragoncoder047:
That's still confusing, then, because if the sub-tree is detached it "doesn't exist" but the objects still have an id.

And re-adding subtrees that had null parents I think is bound to create more problems than it solves, since the physics system will be confused if the object is removed in this way.

mflerackers:
But clearing parent on all would dismantle the sub-tree. So clearing id on all may be better, we can just give them a new id when reattached.

dragoncoder047:
That's probably a task for later down the road.

mflerackers (approved):

PR #856, opened 2025-07-31
fix: ignore destroyed objects when colliding

dragoncoder047:

now that #849 was merged I have started to notice that the area object can ocasionally trigger the 'area() requires the obj to have an id' error when it collides with a destroyed object in the tick immediately after it is destroyed but before it is removed from the physics system. This makes the area component ignore objects that don't exist() which is probably what was intended anyway.

(For example, launch the platformer example, and then stomp on a ghosty.)

### DIFF
--- a/src/ecs/components/physics/area.ts
+++ b/src/ecs/components/physics/area.ts
@@ -267,12 +267,7 @@ export function area(opt: AreaCompOpt = {}): AreaComp {
 
             events.push(
                 this.onCollideUpdate((obj, col) => {
-                    if (!obj.id) {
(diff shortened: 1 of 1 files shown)


mflerackers:
if (!obj.id) {
                        if (!obj.exists()) return;

But isn't exists already testing the id?

dragoncoder047:
It is, but the check only checks the id whereas exists() checks both the id and the parent

mflerackers:
Yeah, but I mean you can do just one check, since an object only exists if it has an id. that throw may go I think.

mflerackers (approved):

PR #857, opened 2025-07-31
feat: expose insertion sort method

### LINKED ISSUE #855: feat: expose public insertionSort method

The sweepandprune uses insertion sort because it's fast on nearly-sorted lists. It's also useful with some custom drawing routines, especially when combined with `get("...", {liveUpdate: true})`. So maybe the insertionSort method can be pulled out to a utils method and then used by the sweepandprune, but also exposed on the kaplay context.

### Any more information?

It would work just like the normal array.sort() method with taking a custom comparator

dragoncoder047:
closes #855

also I deleted the thing about the sapDirection in the changelog cause it doesn't exist (how did it get there? probably my fault...)

### DIFF
--- a/src/math/spatial/sweepandprune.ts
+++ b/src/math/spatial/sweepandprune.ts
@@ -1,6 +1,7 @@
 import type { AreaComp } from "../../ecs/components/physics/area";
 import { isPaused } from "../../ecs/entity/utils";
 import type { GameObj } from "../../types";
+import { insertionSort } from "../sort";
 import { calcTransform } from "../various";
 
 /**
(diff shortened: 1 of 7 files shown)


mflerackers:
Extracting the function to use it elsewhere is OK. But you changed the sweepAndPrune code too. Since the swap and compare are now function calls and no longer inline. I think you should benchmark that first.

mflerackers:
From a few benchmarks I can say that the non-inline versions are only half as fast, so that change will impact the collision phase.

mflerackers:
Mind you, I would do it exactly like that in C++, but Javascript is really bad at optimization, so you can't write nice code in it.

dragoncoder047:
I figured it might have an impact but thought that most good JS engines would automatically inline it because it's "hot" code. I can put it back when I get a chance

src/math/spatial/sweepandprune.ts
```diff
@@ -74,7 +74,7 @@ export class SweepAndPrune {
             edges[0].x = bbox.pos.x;
             edges[1].x = bbox.pos.x + bbox.width;
         }
-        // Insertion sort. This is slow the first time, but faster afterwards as the list is nearly sorted
```
mflerackers, line 77: Maybe don't delete the comment why we use insertion sort, and not the built-in stable sort

PR #862, opened 2025-08-03
docs: add some useful info to KAPLAYOpt.crisp

bobrosoft:
It's improvement in KAPLAYOpt.crisp API reference.

## Summary

- [ ] Changeloged

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -268,7 +268,7 @@ export interface KAPLAYOpt<
      */
     pixelDensity?: number;
     /**
-     * Disable antialias and enable sharp pixel display.
+     * Disable antialias and enable sharp pixel display. If you see rendering artifacts, play with "pixelDensity" and "scale" params (set them to 2, for start).
      */
     crisp?: boolean;
     /**


src/types.ts
```diff
@@ -268,7 +268,7 @@ export interface KAPLAYOpt<
      */
     pixelDensity?: number;
     /**
-     * Disable antialias and enable sharp pixel display.
+     * Disable antialias and enable sharp pixel display. If you see rendering artifacts, play with "pixelDensity" and "scale" params (set them to 2, for start).
```
imaginarny: I wouldn't word it like that, because it depends on use case and game setup. If someone already has FHD dimension and scales it with scale 2 and pixelDensity 2, it will almost definitely crash memory, on phones definitely. 

I guess it's too complex topic for just a note, but I would change it like this:
If you see rendering artifacts, set "pixelDensity" param to `Math.min(devicePixelRatio, 2)` and "scale" to FHD resolution (e.g. 960x540 would need scale 2). Will result in up to 4K.

(Btw, you can add backticks to \`Math.min(devicePixelRatio, 2)\`, should be formatted as code in docs).

imaginarny (changes requested):
Needs to be reworded as it isn't simple like that and could cause issues actually.

imaginarny (approved):
Looks good to me, what do you think @lajbel?

lajbel:
Excellent, run formatter

bobrosoft:
```
> kaplay@4000.0.0-alpha.20 fmt
> dprint fmt

bobr$ git status
On branch origin/patch-1
nothing to commit, working tree clean
```

PR #863, opened 2025-08-03
fix: moving mouse changes input device type

amyspark-ng:
yeah hopefully this is fine


dragoncoder047:
could also do similar for if the gamepad stick is moved although you'll have to implement a deadzone

mflerackers (approved):

PR #866, opened 2025-08-05
fix: also check gl.getProgramInfoLog for errors

dragoncoder047:
This also checks the final program error log and reports any errors there, in case the program was syntactically correct and compiles, but still can't be linked into a runnable shader.

I have not been able to trigger this bug in the kaplayground for some reason... but it nonetheless occurred in my game so I'm fixing it

- [X] Changeloged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ best friend, lajbel, can put the correct version name here
 
 ## [unreleased]
 
+- `loadShader()` now also checks for link errors as well as compile errors and
+  reports them rather than just silently trying to use a borked shader -
+  @dragoncoder047
+
 ## [unreleased] (v3001)
 
 ---
@@ -159,9 +163,6 @@ best friend, lajbel, can put the correct version name here
 - PatrolComp is not going to last waypoint
   ([#734](https://github.com/kaplayjs/kaplay/issues/734)) - @nojaf
 - Fixed non-focused textInput component backspace - @KeSuave
-- `loadShader()` now also checks for link errors as well as compile errors and
-  reports them rather than just silently trying to use a borked shader -
-  @dragoncoder047
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


src/assets/shader.ts
```diff
@@ -100,6 +100,9 @@ export class Shader {
             if (vertError) throw new Error("VERTEX SHADER " + vertError);
             const fragError = gl.getShaderInfoLog(fragShader!);
             if (fragError) throw new Error("FRAGMENT SHADER " + fragError);
+            const linkError = gl.getProgramInfoLog(prog!);
+            if (linkError) throw new Error("LINK ERROR: " + linkError);
+            throw new Error("Unknown shader error (gl.LINK_STATUS was false)");
```
lajbel, line 105: I've added a `throwError` function, would not the best use it?

CHANGELOG.md
```diff
@@ -104,6 +104,9 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - PatrolComp is not going to last waypoint
   ([#734](https://github.com/kaplayjs/kaplay/issues/734)) - @nojaf
 - Fixed non-focused textInput component backspace - @KeSuave
+- `loadShader()` now also checks for link errors as well as compile errors and
```
lajbel: Changelog in unreleased

lajbel (changes requested):

src/assets/shader.ts
dragoncoder047: no, because errors thrown here get caught elsewhere already. for loadShader() (direct from string) it will cause the whole script to crash and black-screen the game, then the error is in the console. for loadShaderURL(), the error is caught by the .then() and results in a failed asset, and then onLoadError() will handle it.

lajbel (approved):

PR #869, opened 2025-08-10
fix: recording now mostly works

dragoncoder047:
ok I think I fixed the record()

also added screenshotBlob() that gives you the screenshot as a blob instead of just a dataURL which can't be downloaded easily like you can now with `downloadBlob("screenshot.png", await screenshotBlob())`

- [X] Changeloged


lajbel:
Awesome, would be cool add in-screen controllers for recording, in debug mode

PR #875, opened 2025-08-13
fix: perf optimization for the shooter game kaplayground example

JSLegendDev:
Someone mentioned under my KAPLAY in 5 mins video that the performance of the shooter example would degrade after 20s if they increased the trash spawn rate to 0.1.

After investigation, I noticed that the offscreen component was not being used, therefore trash game objects would not be destroyed once they fall outside the screen. Since we continually add trash game objects in this game, this would lead to degraded performance over time.

The fix is simple, use the offscreen() component.

### DIFF
--- a/examples/shooter.js
+++ b/examples/shooter.js
@@ -245,6 +245,7 @@ scene("battle", () => {
             pos(rand(0, width()), 0),
             health(OBJ_HEALTH),
             anchor("bot"),
+            offscreen({ destroy: true }),
             "trash",
             "enemy",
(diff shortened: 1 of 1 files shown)


awulkan:
That someone was me. 🙂
Implementing this change didn't fix my performance problems, but it still feels like a good addition to the demo.

mflerackers (changes requested):
Maybe you can remove the old destroy code, and add the collision ignore rules

lajbel:
PR is inactive. Reopen if that changes.

PR #877, opened 2025-08-18
feat: Named animations

mflerackers:


lajbel:
Could you add a Changelog? Maybe explaining how to use this feature so devs can actually use it

PR #878, opened 2025-08-20
Added a rule system for enemy AI

mflerackers:


dragoncoder047:
this isn't exported anywhere so technically it's still unusable by game authors

mflerackers:
I know, but I doubt anyone is going to use it without documentation anyway.

dragoncoder047:
well like an example would help initially, so maybe it could be added to the ghosthunting example or something

mflerackers:
Yes, but I feel like there should be a component for it, but I have no idea yet how that would look.

lajbel:
Shouldn't changelog it if it's not exported

mflerackers:
I'll add an example and export it.

lajbel (approved):

PR #881, opened 2025-08-25
stand-alone state machine class

mflerackers:


src/math/ai/statemachine.ts
```diff
@@ -0,0 +1,125 @@
...
+type ExitCallback = (sm: StateMachine, next: string) => void;
+type TransitionCallback = (sm: StateMachine, from: string, to: string) => void;
+
+type StateCallbacks = {
+    enter?: EnterCallback;
+    update?: UpdateCallback;
+    draw?: DrawCallback;
+    exit?: ExitCallback;
+};
+
```
lajbel, line 13: There's a reason to use Type Aliases instead of just put the type? I think it makes documentation verbose

lajbel (changes requested):

src/math/ai/statemachine.ts
mflerackers: Those are all internal, not exported, so they won't show up in the documentation.

lajbel: Our documentation process export the types even if they're not "exported" because it checks for the references, so they will appear anyway

mflerackers: So how can we make sure the docs don't pick up internal types, functions and classes?

lajbel: But they have to, if no how an user would know what have to pass here in `cb`:

https://github.com/kaplayjs/kaplay/blob/1fd31ea36ed35b2b395329cad434474373d45173/src/math/ai/statemachine.ts#L70-L75

I mean, then the user must have to click `EnterCallback` to know what to pass here, imagine we did this with all our functions that would be a lot of types that you have to click on for knowing them

lajbel:
Awesome work, thank you so much ❤️✨

lajbel (approved):

dragoncoder047:
any plans now to make the state component use one of these internally?

mflerackers:
Probably yes, but I'm first going to add a hierarchical state machine and a pushdown automaton and see whether we give state() more functionality or not. Right now it would only cut some code.
Also note that the StateMachine does not use events, but callbacks, since it is meant to be low level.

PR #883, opened 2025-08-26
fix: add serialize to text component like it should

dragoncoder047:
https://discord.com/channels/883781994583056384/883782079802908772/1409917499784433717

also should any of the other physics (effectors) components be serialized

### DIFF
--- a/src/ecs/components/draw/text.ts
+++ b/src/ecs/components/draw/text.ts
@@ -186,7 +186,7 @@ export function text(t: string, opt: TextCompOpt = {}): TextComp {
     let _width = opt.width ?? 0;
     let _height = 0;
 
-    const obj = {
+    const obj: TextComp = {
         id: "text",
         set text(nt) {
             t = nt;
@@ -197,7 +197,7 @@ export function text(t: string, opt: TextCompOpt = {}): TextComp {
             return t;
         },
         textSize: opt.size ?? DEF_TEXT_SIZE,
-        font: opt.font,
+        font: opt.font!,
         get width() {
             return _width;
         },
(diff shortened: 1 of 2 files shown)


src/ecs/components/draw/text.ts
```diff
@@ -212,11 +212,11 @@ export function text(t: string, opt: TextCompOpt = {}): TextComp {
...
-        align: opt.align,
-        lineSpacing: opt.lineSpacing,
-        letterSpacing: opt.letterSpacing,
-        textTransform: opt.transform,
-        textStyles: opt.styles,
+        align: opt.align!,
+        lineSpacing: opt.lineSpacing!,
+        letterSpacing: opt.letterSpacing!,
+        textTransform: opt.transform!,
+        textStyles: opt.styles!,
```
lajbel, line 219: > In the past we had problems with those !, specify why using on each case (or serie of cases)

Same thing

src/ecs/components/draw/text.ts
```diff
@@ -197,7 +197,7 @@ export function text(t: string, opt: TextCompOpt = {}): TextComp {
             return t;
         },
         textSize: opt.size ?? DEF_TEXT_SIZE,
-        font: opt.font,
+        font: opt.font!,
```
lajbel, line 200: In the past we had problems with those !, specify why using on each case (or serie of cases)

lajbel (changes requested):

src/ecs/components/draw/text.ts
dragoncoder047: This worked before, so it shouldn't be a problem, it's just that typescript complained  about undefined so I added those to shut it up. Maybe I should have changed the TextComp definition since it's clear that it can be undefined and still work?

src/ecs/components/draw/text.ts
lajbel: @dragoncoder047

lajbel: nvm

PR #898, opened 2025-09-22
feat: scrollcam component

itzKiwiSky:
added a new component, the `scrollcam` component implements a simple parallax system inspired by haxeflixel functions like `setScrollFactor`.

the example test also available for further test and verifications.

## Summary

- [x] `scrollcam` component
- [x] test for the component

https://github.com/user-attachments/assets/c1836303-38c3-410a-9d80-c8e38af91b9f


src/ecs/components/transform/camscroll.ts
```diff
@@ -0,0 +1,76 @@
...
+     * @returns The current factor of the object as a {@link Vec2 `Vec2`}
+     */
+    factor: Vec2;
+
+    /**
+     * Set the object scroll factor
+     * @param x number
+     * @param y number
+     */
+    setScrollFactor(x: number, y: number): void;
```
dragoncoder047: should be `...args: Vec2Args` rather than hardcoding x and y

src/ecs/components/transform/camscroll.ts
```diff
@@ -0,0 +1,76 @@
...
+
+
+/**
+ * The {@link camscroll `camscroll()`} component.
+ *
+ * @group Components
+ * @subgroup Component Types
+ */
+export interface CamScrollComp extends Comp 
+{
```
dragoncoder047: run `pnpm fmt`

src/ecs/components/transform/parallax.ts
```diff
@@ -0,0 +1,76 @@
...
+     * @param y number
+     */
+    setScrollFactor(x: number, y: number): void;
+
+    /**
+     * Get the current object scroll factor
+     * @returns The current factor of the object as a {@link Vec2 `Vec2`}
+     */
+    getScrollFactor(): Vec2;
+}
```
dragoncoder047, line 25: why is `basePos` not exposed on the interface

src/ecs/components/transform/camscroll.ts
```diff
@@ -0,0 +1,76 @@
...
+export function scrollcam(x: number = 1, y: number = 1): CamScrollComp 
+{
+    return {
+        id: "scroll",
+        require: ["pos"],
+        factor: new Vec2(x, y),
+        //@ts-ignore
+        basePos: null as Vec2 | null,
+
+        add()
```
dragoncoder047: this add hook does nothing so remove it

dragoncoder047 (changes requested):

src/ecs/components/transform/camscroll.ts
```diff
@@ -34,28 +34,19 @@ export interface CamScrollComp extends Comp
...
-            let self = this as unknown as GameObj<PosComp>;
-            console.log(x, y);
-        },
+        factor: vec2(...args),
+        basePos: vec2(),
 
-        update() 
-        {
-            const obj = this as unknown as GameObj<PosComp> & { basePos: Vec2 | null };
+        update() {
```
dragoncoder047: you can just use `update(this: GameObj<PosComp | CamScrollComp>) {`

src/ecs/components/transform/camscroll.ts
```diff
@@ -64,13 +55,17 @@ export function scrollcam(x: number = 1, y: number = 1): CamScrollComp
             obj.pos.y = obj.basePos.y - cam.y * this.factor.y;
         },
 
-        setScrollFactor(x: number, y: number) { 
+        setScrollFactor(x: number, y: number) {
```
dragoncoder047: also vec2args here

src/ecs/components/transform/camscroll.ts
```diff
@@ -64,13 +55,17 @@ export function scrollcam(x: number = 1, y: number = 1): CamScrollComp
...
             this.factor.x = x;
-            this.factor.y =  y; 
+            this.factor.y = y;
         },
 
-        getScrollFactor() { return this.factor.clone() },
+        getScrollFactor() {
```
dragoncoder047: also are thes necessary when you can access the factor property (it's not private / in the closure)

src/core/contextType.ts
```diff
@@ -1442,6 +1443,13 @@ export interface KAPLAYCtx {
      * @subgroup Rendering
      */
     fixed(fixed?: boolean): FixedComp;
+    /**
+     * Make the object scrolls with camera based on a factor
+     * @param x number
+     * @param y number
+     * @returns the camera scroll comp
+     */
+    scrollcam(x: number, y: number): CamScrollComp;
```
dragoncoder047: should still be Vec2Args here?

dragoncoder047 (changes requested):

src/core/contextType.ts
itzKiwiSky: is bug??

itzKiwiSky: now should be fixed

src/ecs/components/transform/camscroll.ts
```diff
@@ -0,0 +1,52 @@
...
+            if (!this.basePos) {
+                this.basePos = this.pos.clone();
+            }
+
+            this.pos.x = this.basePos.x - cam.x * this.factor.x;
+            this.pos.y = this.basePos.y - cam.y * this.factor.y;
+        },
+
+        getScroll() {
+            return this.factor.clone();
```
mflerackers: Why have getScroll if factor is public? Was factor supposed to be a closure variable instead maybe?

mflerackers: You probably meant
```ts
    let factor = vec2(...args);
    let basePos = vec2();
    return {
        id: "scroll",
        require: ["pos"],
```
So they are inaccessible from outside

mflerackers (changes requested):

mflerackers (approved):

src/core/contextType.ts
```diff
@@ -1478,6 +1480,13 @@ export interface KAPLAYCtx {
      * @subgroup Rendering
      */
     fixed(fixed?: boolean): FixedComp;
+
+    /**
+     * Make the object scrolls with camera based on a factor
```
dragoncoder047, line 1504: I would also add a mention of "parallax scrolling" since that's what this component also does.

it might be more intuitive to rename the component as well to `parallax()` but that's up to you

src/ecs/components/transform/camscroll.ts
```diff
@@ -0,0 +1,46 @@
...
+    return {
+        id: "scroll",
+        require: ["pos"],
+        factor: vec2(...args),
+        basePos: vec2(),
+
+        update(this: GameObj<PosComp | CamScrollComp>) {
+            const cam = getCamPos();
+
+            if (!this.basePos) {
```
dragoncoder047: this line will never work because you're setting it intially to the zero vector on line 29 which is an object and considered truthy

dragoncoder047 (changes requested):

src/ecs/components/transform/camscroll.ts
```diff
@@ -0,0 +1,46 @@
...
+     * The current factor for object scroll
+     *
+     * @returns The current factor of the object as a {@link Vec2 `Vec2`}
+     */
+    factor: Vec2;
+
+    basePos: Vec2;
+}
+
+export function scrollcam(...args: Vec2Args): CamScrollComp {
```
dragoncoder047: another idea: instead of the vanishing point always being the camera position, you could have another argument that would change the vanishing point relative to the screen rectangle to change the perspective. So like for `vec2(0, -1/3)` would mean the vanishing point is in the center but 1/6 of the screen height below the camera.

dragoncoder047 (changes requested):

src/ecs/components/transform/camscroll.ts
itzKiwiSky: can you explain it a bit better? is not making much sense to me.

itzKiwiSky: wait, you're asking for a way to set the point on any offset from the camera?
this could be good, I can try

itzKiwiSky:
I made some changes to the parallax component because the old code has been broken, the effect has not being applied so, I used the literal of MoveBy to fix it

src/ecs/components/transform/parallax.ts
```diff
@@ -38,7 +38,7 @@ export function parallax(...args: Vec2Args): ParallaxComp {
             }
 
             const delta: Vec2 = cam.sub(this.lastCam);
-            this.moveBy(
+            this.pos = this.pos.add(
```
mflerackers: moveBy does exactly this. You're saying it isn't working?

mflerackers (changes requested):

src/ecs/components/transform/parallax.ts
itzKiwiSky: I thought this was about organization, I will return back to the original version instead

mflerackers: We might remove the pos setter/getter in the future and have only moveTo, moveBy and getPos. So it's better to leave moveBy.

itzKiwiSky: yea I notice I "can't" set the pos directly, that's why broke the old version.

itzKiwiSky: yea I notice I "can't" set the pos directly, that's why broke the old version.

mflerackers: You can, just not pos.x or pos.y separately, you need to replace the entire pos. That's why we're thinking of removing that property.

itzKiwiSky: yea make sense, will be for good if those properties be private, only internal access

amyspark-ng:
@itzKiwiSky missing changelog btw i think

amyspark-ng:
@itzKiwiSky what's the state on this?

itzKiwiSky:
Is better to left on draft, or probably close all togheter.

amyspark-ng:
not going to continue it?

itzKiwiSky:
like dont merge and just refuse everything idk

itzKiwiSky:
not plan to continue, not too much time to work on it, sorry, when I got some free time I want to check more about my old pr, the pushScene and popScene

amyspark-ng:
alright, will be closed for now

PR #916, opened 2025-10-17
fix: examples

dragoncoder047:
some of them were bugged so I fixed them

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,7 +104,12 @@ best friend, lajbel, can put the correct version name here
   });
   ```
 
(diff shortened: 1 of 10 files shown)


examples/health.js
```diff
@@ -69,7 +69,7 @@ bean.onKeyPress("space", () => {
     ]);
 
     bullet.onCollide(() => {
-        bullet.destroy();
```
lajbel, line 72: why

dragoncoder047: the old way it was being destroyed *before* the other onCollide (which dealt damage to the zombeam), new way it destroys 1 frame later and it works

PR #934, opened 2025-10-28
fix: use app scope so constraints get processed always

dragoncoder047:
changing scenes broke the constraint system's events becuase the default is scene scope but it wouldn't reinstall

### DIFF
--- a/examples/fps.js
+++ b/examples/fps.js
@@ -1,0 +1,5 @@
+kaplay({
+    maxFPS: 70,
+});
+
+debug.inspect = true;
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -378,11 +378,16 @@ export const initApp = (
                         fixedUpdate();
                     }
                     state.restDt = fixedAccumulatedDt;
-                    state.dt = accumulatedDt;
+                    state.dt = desiredDt;
                     state.time += dt();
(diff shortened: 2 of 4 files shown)


src/ecs/components/transform/constraint.ts
```diff
@@ -174,25 +175,29 @@ function installSystem() {
     if (systemInstalled) return;
     systemInstalled = true;
     // TODO: use a live query for this
+    // also need to fix live queries to work with scene/app scopes
     const constraints: Set<GameObj<Constraint>> = new Set();
-    onAdd(obj => {
+    _k.k.scene.onAdd(obj => {
         if (obj.has("constraint")) {
```
lajbel, line 181: use _k.k.app because you can change default scope, also onAdd alone is just scene.onAdd

lajbel (changes requested):

src/ecs/components/transform/constraint.ts
dragoncoder047: i tried that the first time and it didn't work fsr

dragoncoder047:
mm I think I'll let you @lajbel figure this out along with the other events stuff. seems too complicated at first glance for met to be able to solve it quickly

PR #944, opened 2025-11-08
feat: Sprite outline tool

milosilo-dev:
It provides an outline tool that fits a polygon to a sprite image automatically. This can be used alongside an area component to give an automatic collider to sprites in Kaplay.
## Summary
Provides a function called `getSpriteOutline` that can be used in addition with an area shape to make a custom collider for a sprite image.
- [ ] added a function to `math.ts` called `getSpriteOutline` which returns the polygon shape from a provided sprite
- [ ] added a new file in the maths directory called `getImageOutline.ts`, which holds the logic for the marching square's algorithm.
- [ ] Added code to `context.ts` and `contextType.ts` which exports the function, so the user can use it
This is my first pull request, so if I made any mistakes, please contact me about them and I will be happy to attempt to resolve any issues.
- [ ] Added a new example called Automatic Collider to show how to use the library

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,13 +20,8 @@ best friend, lajbel, can put the correct version name here
 
 ### Added
 
-- New example called `automaticCollider.js` that automatically creates colliders
-  for a given sprite
-- Exports function called `getSpriteOutline` that takes a sprite asset and
+- Added `getSpriteOutline()` that takes a sprite asset and
   returns a polygon showing the outline.
-
-### Added
-
 - Added parameterized formatting tags like `"[color=red]Red text![/color]"` in
   `CharTransformFunc` for more powerful text formatting options -
   @dragoncoder047


mflerackers:
You might use something like Ramer–Douglas–Peucker to reduce the amount of polygon points though.

milosilo-dev:
> You might use something like Ramer–Douglas–Peucker to reduce the amount of polygon points though.

That's a very good point because the polygons can become very complex on large sprites, I will look into implementing the algorithm that you mentioned.

dragoncoder047:
Oooh amazing! I have previously been doing this with a shader but being able to get a polygon is nice!

I think if you're using it to create colliders you will need to implement some decimation algorithm like RDP since KAPLAY's current physics collision algorithm can't handle concave polygons nor compound colliders so a strict pixel tracing approach will cause weird issues.

That being said I have experimented with vectorization algorithms in the past and I have a couple of algorithms that aren't RDP, but still do the job, and work better for pixel sprites, so maybe you could implement RDP *but* also add an option to turn it off and allow the user to simplify the polygon themselves.

milosilo-dev:
I have added RDP now, look over my implementation, but I basically translated [this](https://medium.com/@indemfeld/the-ramer-douglas-peucker-algorithm-d542807093e7) one from python to JS.

examples/automaticCollider.js
```diff
@@ -0,0 +1,44 @@
...
+    });
+});
+
+var poly;
+onLoad(() => {
+    poly = getSpriteOutline("apple");
+
+    apple = add([
+        sprite("apple"),
+        area(getSpriteOutline("apple")),
```
mflerackers: I think this should be area(poly), no?

src/math/getImageOutline.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (open.length > 1 && open[0].dist(open[open.length - 1]) < 1e-3) {
+        open.pop();
+    }
+    const simplified = RDP(open, epsilon);
+
+    simplified.push(simplified[0]);
+    return simplified;
+}
+
+function getDistance(c: Vec2, a: Vec2, b: Vec2): number {
```
mflerackers, line 81: This needs some comment like
```ts
/**
 * Calculates the perpendicular distance between c and the line from a to b.
 */
```

src/math/getImageOutline.ts
```diff
@@ -0,0 +1,124 @@
...
+    const A = b.y - a.y;
+    const B = a.x - b.x;
+    const C = b.x * a.y - a.x * b.y;
+
+    return Math.abs(A * c.x + B * c.y + C) / Math.sqrt(A * A + B * B);
+}
+
+function RDP(points: Vec2[], epsilon: number): Vec2[] {
```
mflerackers, line 92: Maybe add a comment that it implements Ramer–Douglas–Peucker, RDP is not really descriptive.

milosilo-dev: I have added the comment

src/math/getImageOutline.ts
milosilo-dev: Added the comment

examples/automaticCollider.js
milosilo-dev: Yup, you are right!

mflerackers (approved):

src/math/getImageOutline.ts
```diff
@@ -0,0 +1,130 @@
+import { Vec2 } from "./Vec2";
+
+export function trace_region(
```
lajbel: so x but please use it as `traceRegion`

examples/automaticCollider.js
```diff
@@ -0,0 +1,44 @@
...
+loadSprite("apple", "sprites/apple.png");
+loadSprite("tga", "sprites/tga.png");
+
+onDraw("gm", (gm) => {
+    drawCircle({
+        radius: 1,
+    });
+});
+
+var poly;
```
lajbel: change for let please

lajbel (changes requested):

src/math/getImageOutline.ts
milosilo-dev: Done

examples/automaticCollider.js
milosilo-dev: Done

lajbel:
Awesome work with documenting, and the code looks nice. May I ask if this is how it's supposed to work?

lajbel:
image

milosilo-dev:
> image

Could you tell me what code you're running for this, Is it the example that I made or is it something else?

milosilo-dev:
Never mind I was able to replicate the issue, I will look into it when I get the chance

src/math/getImageOutline.ts
```diff
@@ -0,0 +1,118 @@
...
+        new Vec2(-1, -1),
+        new Vec2(0, -1),
+        new Vec2(1, -1),
+    ];
+
+    // Find a starting pixel in the region
+    let start: Vec2 | null = null;
+    for (let y = 0; y < height && !start; y++) {
+        for (let x = 0; x < width && !start; x++) {
+            if (isInRegion(x, y)) start = new Vec2(x, y);
```
mflerackers: Shouldn't you break once you found one?

milosilo-dev: Yes not sure how I missed that!

mflerackers (approved):

lajbel (approved):

lajbel:
@milosilo-dev Awesome, could we get a changelog for it?

PR #950, opened 2025-11-15
fix: Render Narrow/Tall Fonts Without Cropping Them

anthonygood:
* Fixes text cropping for narrow fonts
* Adds a `/textTall` test and two specimen fonts which exhibit the bug

The root of the problem [as discussed here](https://discord.com/channels/883781994583056384/883793204179251311/1439236351085121588) is that the `textBaseline = 'top'` refers not to the top of the text's bounding box but [the em-over baseline](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-textbaseline-top). Therefore some characters can and do ascend beyond this point, causing the cropping on tall/narrow fonts.

### Before
image

### After
image

## Summary

- [ ] Changeloged

### DIFF
--- a/src/gfx/formatText.ts
+++ b/src/gfx/formatText.ts
@@ -213,14 +213,16 @@ function updateFontAtlas(font: FontData | string, ch: string) {
         c2d.fillStyle = "#ffffff";
 
         if (atlas.maxActualBoundingBoxAscent === 0) {
-            atlas.maxActualBoundingBoxAscent = c2d.measureText(allChars()).actualBoundingBoxAscent;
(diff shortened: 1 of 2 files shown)


lajbel:
Very good, thank you so much! We tried like 10000 times to fix this bug

lajbel:
Oops @anthonygood could you add you in the changelog? Just open a new PR

anthonygood:
Sorry @lajbel what do you mean? Happy to oblige, though I’m on holiday until next Tuesday! Will submit with any additional info when I’m back if that’s ok.

lajbel:
NVM, have a great holiday!

PR #955, opened 2025-11-24
Broad phase configuration, vertical Sweep and Prune and Quadtree structure

mflerackers:
## Summary

- [ ] Changeloged

### DIFF
--- a/src/ecs/components/physics/area.ts
+++ b/src/ecs/components/physics/area.ts
@@ -396,9 +396,10 @@ export function area(opt: AreaCompOpt = {}): AreaComp {
             }
             else {
                 if (!otherOrTag.id) {
-                    throw new Error(
+                    /*throw new Error(
                         "isColliding() requires the object to have an id",
-                    );
+                    );*/
+                    return false;
                 }
                 return Boolean(colliding[otherOrTag.id]);
             }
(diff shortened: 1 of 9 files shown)


src/ecs/components/physics/area.ts
```diff
@@ -396,9 +396,10 @@ export function area(opt: AreaCompOpt = {}): AreaComp {
             }
             else {
                 if (!otherOrTag.id) {
-                    throw new Error(
+                    /*throw new Error(
                         "isColliding() requires the object to have an id",
```
lajbel, line 400: goody good

src/ecs/systems/createCollisionSystem.ts
```diff
@@ -49,48 +70,47 @@ export const createCollisionSystem = ({ narrow = "gjk" } = {}) => {
...
-                    sap.add(obj as GameObj<AreaComp>);
+                    broadPhaseIntersection.add(obj as GameObj<AreaComp>);
                 }
             }
         }
 
-        sap.update();
-        for (const [obj1, obj2] of sap) {
+        broadPhaseIntersection.update();
+        for (const [obj1, obj2] of broadPhaseIntersection) {
```
dragoncoder047, line 113: Instead of allocating an iterator couldn't you just make a method that takes a callback and calls it on each pair? I think this reduces memory allocation a good bit last time I tested.

dragoncoder047 (changes requested):

src/ecs/systems/createCollisionSystem.ts
mflerackers: Yes. It was written like this because it looked more elegant, and the iteration code looked better also, when using iterators.

PR #960, opened 2025-11-27
fix: don't call onUse() or onTag() while make() is running

dragoncoder047:
@lajbel please check this for hidden gotchas


lajbel:
Looks fine but will add some tests

lajbel (approved):

PR #971, opened 2025-12-04
fix: use music volume given in options

dragoncoder047:
@Stanko found this, i'm not going to make him file an issue for everything ;)

## Summary

- [X] Changeloged


lajbel:
silly bug! ty

PR #976, opened 2025-12-16
feat: add multiturn tracking to rotation constraint; rename scale to ratio

dragoncoder047:
title

## Summary

- [X] Changeloged


CHANGELOG.md
```diff
@@ -33,8 +35,13 @@ best friend, lajbel, can put the correct version name here
 - The grid broadphase has been rewritten for performance. @mflerackers
 - Global `retrieve()` method to get the objects with area within a certain
   rectangle. @mflerackers
+- The rotation constraint's `scale` option has been **renamed** to `ratio` -
```
lajbel: you should mark it as a breaking change ig?

lajbel (changes requested):

mflerackers (approved):

lajbel:
Please tag me when the merge conflicts are fixed. I'll merge

dragoncoder047:
looks like no conflicts i think

lajbel:
What is exactly the point of this PR?

dragoncoder047:
The point is to make it so that two objects with rotation constraint between them is able to work when ~~scale~~ ratio is less than 1, normally since the source's angle is 0-360 when it wraps back from 360 to 0 the destination object (the one the constraint is attached to) is going to suddenly jump back to being at its zero position, but this code automatically tracks that boundary jump and compensates so that the destination's rotation is continuous.

Also, since now the constraint actually behaves like gears/pulleys, it just makes sense to rename scale to ratio.

amyspark-ng:
@lajbel

amyspark-ng:
@dragoncoder047 make sure this is properly changelogged btw

lajbel:
Would you show me an example with videos? @dragoncoder047

amyspark-ng:
requested changes, marking as waiting for author

PR #977, opened 2025-12-17
feat(scopes): Add missing events for scopes feature [add/destroy/use/unuse/tag/untag]

lajbel:
## Summary

- [x] Changeloged
- [x] Normalize `on()`
- [x] Tested


src/api/eventHandlers.ts
```diff
@@ -0,0 +1,227 @@
...
+): KEventController => {
+    let paused = false;
+    let obj2Handler = new Map<GameObj, KEventController>();
+
+    const handleNew = (obj: GameObj) => {
+        const ec = obj.on(event, (...args) => {
+            cb(obj, ...<TupleWithoutFirst<GameObjEvents[Ev]>> args);
+        });
+        ec.paused = paused;
+        if (obj2Handler.has(obj)) obj2Handler.get(obj)!.cancel();
```
mflerackers: Can't you do something like

```ts
let h;
if (h = obj2Handler.get(obj)) h.cancel();
```

So you only do one lookup?

src/api/eventHandlers.ts
```diff
@@ -0,0 +1,227 @@
...
+        if (obj.is(tag)) handleNew(obj);
+    });
+    const ecOnUntag = _k.appScope.onUntag((obj, oldTag) => {
+        if (oldTag === tag) {
+            const ec = obj2Handler.get(obj)!;
+            ec.cancel();
+            obj2Handler.delete(obj);
+        }
+    });
+
```
mflerackers, line 58: No onDestroy needed?

dragoncoder047: the event was attached to the object on line 32 so no

src/ecs/components/transform/offscreen.ts
```diff
@@ -124,7 +125,9 @@ export function offscreen(opt: OffScreenCompOpt = {}): OffScreenComp {
             return this.on("enterView", action);
         },
         add(this: GameObj<OffScreenComp>) {
-            if (opt.pause && opt.unpause) onUpdate(() => check(this));
+            if (opt.pause && opt.unpause) {
```
dragoncoder047, line 128: @lajbel you could fix that by having the object re-setup the events if it gets a `sceneEnter` event, put line 129 in an arrow function, call it once to start, and then attach it in an `on("sceneEnter")` event on `this`

src/api/eventHandlers.ts
lajbel: good catch

PR #979, opened 2025-12-19
feat: Introduces sensors

mflerackers:
## Summary

- [ ] Changeloged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,28 +16,42 @@ Hey, KAPLAY Dev, you must changelog here, in unreleased, so later your
 best friend, lajbel, can put the correct version name here
 -->
 
-### Changed
+## [unreleased]
 
-- Added the `fakeMouseMove` event in `FakeMouseComp`, it will triggers when you
-  move the object.
-- Transforms are now only recalculated when needed. Thus static objects no
-  longer increase computation in the transform phase. @mflerackers
-- Areas are now only recalculated when the area settings or (optional)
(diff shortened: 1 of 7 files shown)


src/math/spatial/hashgrid.ts
```diff
@@ -294,3 +296,7 @@ export class HashGrid {
             * this.cellSize;
     }
 }
+
```
dragoncoder047, line 303: why is this function duplicated 3 times on each file? you could just put it in math/spatial/index.ts and export it

mflerackers: It's intentionally, so that it bothers me, since I need to change that. I'm currently thinking how though. Since it filters what counts as valid collide-able, I'm thinking whether it shouldn't be some setting. For one it makes it easier to get backwards compatibility (ignore sensor and body check), and maybe users might use it (but still thinking of good enough scenarios).

CHANGELOG.md
```diff
@@ -34,6 +35,8 @@ best friend, lajbel, can put the correct version name here
 - The grid broadphase has been rewritten for performance. @mflerackers
 - Global `retrieve()` method to get the objects with area within a certain
   rectangle. @mflerackers
+- Added isSensor. Areas without body or isSensor will no longer be eligible for
```
lajbel:
```
- Added `AreaCompOpt.isSensor`. Areas without body or is sensor will no longer be eligible for collisions - @mflerackers
```

lajbel (approved):

PR #986, opened 2025-12-30
refactor: Convert hover handler to a system

mflerackers:
## Summary

- [ ] Changeloged

### DIFF
--- a/src/ecs/components/physics/area.ts
+++ b/src/ecs/components/physics/area.ts
@@ -83,8 +83,8 @@ function startClickHandler() {
 
 function hoverHandler() {
     let oldObjects: Set<GameObj<AreaComp>> = new Set();
-    return (pos: Vec2) => {
-        const p = toWorld(pos);
+    // p Should be world coordinates
+    return (p: Vec2) => {
         const newObjects: Set<GameObj<AreaComp>> = new Set();
 
         _k.game.retrieve(new Rect(p.sub(1, 1), 3, 3), obj => {
@@ -124,7 +124,7 @@ function startHoverSystem() {
 
     system("hover", () => {
         if (_k.game.fakeMouse) {
-            fakeMouseHover(_k.game.fakeMouse.pos);
+            fakeMouseHover(_k.game.fakeMouse.worldPos()!);
             return;
         }
 


src/ecs/components/physics/area.ts
```diff
@@ -110,6 +112,26 @@ function startHoverHandler() {
...
+function startHoverSystem() {
+    if (systemInstalled) return;
+    systemInstalled = true;
+
+    const mouseHover = hoverHandler();
+    const fakeMouseHover = hoverHandler();
+
+    system("hover", () => {
+        if (_k.game.fakeMouse) {
+            mouseHover(toWorld(_k.game.fakeMouse.pos));
```
dragoncoder047: Shouldn't this be fakeMouseHover or is it intentionally wrong

Also, return if fake mouse exists?

mflerackers: right

mflerackers: Can't you use fakeMouse and Mouse together?

mflerackers: They have their own closure because I thought they can be used simultaneously.

dragoncoder047: Idk, I think the intention was that if the game used fake mouse then the fake mouse object will move to always follow the real mouse, so the hover handler should not run twice in that case. 

Also, the hover handler is given the world position, and toWorld converts from screen space to world space, but the fakeMouse.pos is already world space right? At least it won't be screen space if the camera transform is not the identity

mflerackers: right, I was actually waiting on lajbel to fix the fakemouse part, as I never use fakemouse.

mflerackers: Wait, can fakemouse be attached to another object, because in that case I need the word position.

dragoncoder047: I would think using _k.game.fakeMouse.worldPos() would be the best move

(since the fake mouse object doesn't have to be on the top level)

mflerackers: Does the fakemouse api actually do this? Or is it bugged if it is not attached to the root?

dragoncoder047: https://github.com/kaplayjs/kaplay/blob/master/src/ecs/components/misc/fakeMouse.ts#L79-L81

dragoncoder047: It should

mflerackers:
```ts
update(this: FakeMouse) {
            if (this.pos.x !== lastPos.x || this.pos.y !== lastPos.y) {
                this.trigger("fakeMouseMove", this.pos.clone()); // TODO: Create a deltaPos
            }
```

It doesn't pass world coordinates to its trigger, so if it is not in root, it's not working as expected.

dragoncoder047: The fakeMouseMove event is new, not surprising that it's wrong in places, but also good cause it's unlikely someone is using this extremely unfinished branch just for that event

mflerackers: It's not in this branch, it's been there since fakemouse.

mflerackers: It's not in this branch, it's been there since fakemouse.

PR #988, opened 2025-12-30
fix: input events now retain their paused state

### LINKED ISSUE #891: bug: input events on game obj don't maintain user set state when object pause and unpause
### Bug Description

make an input event on an object and pause it. but when the object itself unpauses, the event is enabled.

the event should behave as though it's a pseudo child of the object, so if the object is paused OR the event itself is paused, the event doesn't run, but if the object is unpaused the event can still remain paused.

### Version

master

### Playground Link

https://play.kaplayjs.com/?code=eJzLTizISazU0LTmSs7PKy5RqFCwVUhMSdGIjoULpQKFKvTy87xTK13yy%2FM0NDQVbO0UqrkUgCAlNak0XS8nP11DKTu1UiEFKK8E1FgLxKl6BYmlxakpQN0lRaWp1lwVeATSEnOKgSII42C6gQYBACUWMyc%3D&version=master

### Extra information

if this worked like a stack, this could be useful to add a characterController() component that handles input interception popups, so that when the popup opens it pauses the event controllers from the previous level, and enables the ones for the popup, and then on close it does the reverse

### Summary

- [ ] Fixed in v4000
- [ ] Fixed in v3001

dragoncoder047:
closes #891

* [X] Changeloged

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,6 @@ name: Test KAPLAY
 
 on:
   push:
-    branches: ["master"]
 
 jobs:
   check-format:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,16 @@ best friend, lajbel, can put the correct version name here
 
 - Added `AreaCompOpt.isSensor`. Areas without body or is sensor will no longer
   be eligible for collisions - @mflerackers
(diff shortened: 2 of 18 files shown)


lajbel (approved):

lajbel:
LGTM, let's see what's MF's opinon

lajbel:
I think MF is overloaded, let's merge.

PR #991, opened 2026-01-01
fix: fix for loop syntax to use 'of' instead of 'in'

benhuangbmj:
Since `Object.keys(opt.tiles)` is used as the iterable in the for loop, `in` will produce the index thus create an error in `opt.tiles[key]`.

## Summary
Corrected the for-loop in the `level` function to use `for...of` instead of `for...in` when iterating over `Object.keys(opt.tiles)`, preventing potential issues with key enumeration.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,6 @@ name: Test KAPLAY
 
 on:
   push:
-    branches: ["master"]
 
 jobs:
   check-format:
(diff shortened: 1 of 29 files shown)


lajbel:
Great! Thank you. Could you add it to the changelog? Remember credit yourself.

benhuangbmj:
> Great! Thank you. Could you add it to the changelog? Remember credit yourself.

Thank you for the instruction. I added it to the changelog.

lajbel:
TYSM

PR #1005, opened 2026-01-12
example: new cool sokoban

amyspark-ng:
new AWESOME sokoban

still missing some sounds


examples/sokoban.js
```diff
@@ -8,74 +8,468 @@
...
+
+// Will store the direction of our movements and if we pushed a box
+let undoStack = [];
+
+let level;
+let player;
+let moveCounter;
+
+// Whether any object in an array has a certain tag
 const hasTag = (objs, tag) => objs.findIndex(obj => obj.is(tag)) !== -1;
```
dragoncoder047:
```suggestion
const hasTag = (objs, tag) => objs.some(obj => obj.is(tag));
```

examples/sokoban.js
```diff
@@ -8,74 +8,468 @@
...
+    level.children.forEach((child, idx) => {
+        child.scale = vec2(0);
+        tween(
+            child.scale,
+            vec2(1),
+            0.07 * idx,
+            (p) => child.scale = p,
+            easings.easeOutBack,
+        ).onEnd(() => {
+            canMove = true;
```
dragoncoder047: shouldn't this only set canMove when the last tween finishes, not the first?

dragoncoder047 (approved):

examples/sokoban.js
```diff
@@ -8,74 +8,471 @@
...
+        opacity(),
+    ]);
+
+    // We go through every tile in the level and do a little animation to scale them in
+    level.children.forEach((child, idx) => {
+        child.scale = vec2(0);
+        tween(
+            child.scale,
+            vec2(1),
+            0.07 * idx,
```
dragoncoder047, line 316: shouldn't this be wait 0.07*idx and then start the tween for a fixed duration? when the last objects' tweens get towards the end, the tail of the easeOutBack makes it look like it's done but it won't let you move

examples/sokoban.js
```diff
@@ -8,74 +8,471 @@
...
+        else if (button == "up") move(vec2(0, -1));
+        else if (button == "down") move(vec2(0, 1));
+    });
+
+    // The undo action
+    onButtonPress("undo", () => {
+        // If there's no more movements to undo return
+        if (undoStack.length === 0) return;
+
+        undos++;
```
dragoncoder047, line 352: shouldn't this also decrement moves and update the moves counter

amyspark-ng: "Cause if you just do a mistake cause clumsy, you are punished.. but, at the same time, it would be unfair if you decreased as if nothing happened and you solved it perfectly.. so not doing anything is probably a middle ground? Idk, we dont have scoring in place, everything could change and make sense if also time was tracked"
> Misa

examples/sokoban.js
imaginarny: Even if the logic in code was an oversight, the idea of waiting until the last tween is finished was not a good idea in the first place for this particular case (as you realized after in the next comment). The tween used has no indication of you being unable to move, as tiles are shown almost immediately and there is just a visual effect going on a bit. So the single 0.07 wait as was the result of the initial unintentional code, was long enough.

dragoncoder047: 70 milliseconds is so short that it's basically 0, so i guess just remove it and have canMove be initially set to true?

dragoncoder047:
@amyspark-ng a good resource for sounds that I like is [ZzFX](https://killedbyapixel.github.io/ZzFX/) and you can go there and then hit the `Save Wav` button on the bottom to save it to the resources folder here

amyspark-ng:
right it seems very good, i should try it instead of jsfxr the next time, even though it doesn't have sliders

dragoncoder047:
also I have the plugin for it in the #plugins in the discord

examples/sokoban.js
```diff
@@ -8,74 +8,479 @@
...
+            : vec4(0, 0, 0, 0.15);
+    }
+`,
+);
+
+/*
+Sokoban is a puzzle game where
+We have to place all the boxes in a sensor to continue the level
+We'll code it according to this logic
+*/
```
imaginarny: Maybe should move it above the code so it can seen in the beginning of the file in kaplayground.

examples/sokoban.js
```diff
@@ -8,74 +8,479 @@
...
+
+// Move an object in a certain direction
 const moveObj = (obj, dir) => {
     if (dir.x == 1) obj.moveRight();
     if (dir.x == -1) obj.moveLeft();
     if (dir.y == 1) obj.moveDown();
     if (dir.y == -1) obj.moveUp();
 };
 
+// The "update" function in our game that runs everytime we make a move
```
imaginarny:
```suggestion
// The "update" function in our game that runs every time we make a move
```

examples/sokoban.js
```diff
@@ -8,74 +8,479 @@
...
 const move = (dir) => {
-    const moveTo = player.tilePos.add(dir);
-    const occupant = level.getAt(moveTo);
+    // The new position the player will be in
+    const playerNewPos = player.tilePos.add(dir);
+
+    // The objects that exists at the position the player is going to move to
```
imaginarny:
```suggestion
    // The objects that exist at the position the player is going to move to
```

examples/sokoban.js
```diff
@@ -8,74 +8,479 @@
...
+    const occupants = level.getAt(playerNewPos).filter((obj) =>
+        !obj.is("sensor")
+    );
+
+    // If occupants is a wall we simply return and don't do anything else
```
imaginarny:
```suggestion
    // If occupants include a wall we simply return and don't do anything else
```

examples/sokoban.js
```diff
@@ -8,74 +8,479 @@
...
+    // If the occupants include a box
+    let box = null;
+    if (hasTag(occupants, "box")) {
+        box = occupants[0];
+
+        // The occupants at the position the box would get moved to
+        // (The same direction you moved to)
+        const boxOccupants = level.getAt(occupants[0].tilePos.add(dir));
+
+        // If there's no occupants or it's a sensor and there isn't a box at this place
```
imaginarny:
```suggestion
        // If there're no occupants, or if there's a sensor and no box at this place
```

examples/sokoban.js
```diff
@@ -8,74 +8,479 @@
...
+        // You should be able to actually push the box
+        if (
+            !boxOccupants[0]
+            || boxOccupants[0].is("sensor") && !hasTag(boxOccupants, "box")
+        ) {
+            // Push the box in the same direction and do a little tween
+            play("movebox", { detune: -rand(-100, 0) });
```
imaginarny: Why not like this? 🙂 
```suggestion
            play("movebox", { detune: rand(0, 100) });
```
It's there one more time, line ~369

examples/sokoban.js
```diff
@@ -8,74 +8,479 @@
...
+            moveObj(box, dir);
+            tween(
+                vec2(1.2),
+                vec2(1),
+                0.12,
+                (p) => box.scale = p,
+                easings.easeOutQuad,
+            );
+        }
+        // If you shouldn't be able to push the box, tween them to signal the push
```
imaginarny: This comment somewhat contradicts itself so maybe:
```suggestion
        // If you shouldn't be able to push the box, but can keep the visual feedback
```

examples/sokoban.js
```diff
@@ -8,74 +8,479 @@
...
+                0.12,
+                (p) => box.scale = p,
+                easings.easeOutQuad,
+            );
             return;
         }
 
-        moveObj(occupant[0], dir);
+        // If the box is in a sensor, make it sturdy to show that you can't move it anymore
+        // (Unless you undo)
```
imaginarny: This is outdated. You can move it out of sensor, so:

```suggestion
        // If the box is in a sensor, change its sprite to indicate it's in place
```

examples/sokoban.js
```diff
@@ -8,74 +8,479 @@
...
-onKeyPress("s", () => {
-    move(vec2(0, 1));
+scene("win", () => {
+    // Checkered background
+    play("explode");
+
+    add([
+        rect(width(), height()),
+        shader("checker", { tileSize }),
+    ]);
```
imaginarny: Either `// Checkered background` or `play("explode");` is in the wrong place.

examples/sokoban.js
```diff
@@ -8,74 +8,479 @@
...
+            else move.box.sprite = "box";
+        }
+    });
 });
 
-onKeyPress("s", () => {
-    move(vec2(0, 1));
+scene("win", () => {
+    // Checkered background
+    play("explode");
```
imaginarny: For me the explosion is at the much higher volume than the rest sounds so might lower it. Do you hear the difference?

examples/sokoban.js
```diff
@@ -8,74 +8,479 @@
...
+                easings.easeOutQuad,
+            );
             return;
         }
 
-        moveObj(occupant[0], dir);
+        // If the box is in a sensor, make it sturdy to show that you can't move it anymore
+        // (Unless you undo)
+        if (isBoxInSensor(box)) box.sprite = "sturdybox";
+        else box.sprite = "box";
```
imaginarny:
```suggestion
        box.sprite = isBoxInSensor(box) ? "sturdybox" : "box";
```

examples/sokoban.js
```diff
@@ -8,74 +8,479 @@
...
+            tween(
+                vec2(0.8),
+                vec2(1),
+                0.15,
+                (p) => move.box.scale = p,
+                easings.easeOutQuad,
+            );
+            // We re-check if it was sturdy, if it's not sturdy anymore we change the sprite back
+            if (isBoxInSensor(move.box)) move.box.sprite = "sturdybox";
+            else move.box.sprite = "box";
```
imaginarny:
```suggestion
            // We re-check if it was in a sensor and set the sprite accordingly
            move.box.sprite = isBoxInSensor(move.box) ? "sturdybox" : "box";
```

amyspark-ng:
sorry!! i hadn't really thought most of these and a bunch of honest mistakes, hope it's fine now

lajbel:
Besides code review, I think some things can be easily improved game wise:

- Create a main menu because the win scene is shown at start
- Add the movements count in the win scene

imaginarny:
Sry, the last review took me an hour plus but I'm 120% focusing on finishing kaplayground update so I'll check after.

> - Create a main menu because the win scene is shown at start

@lajbel I think @amyspark-ng was just debugging it and forgot to revert scene to `"game"` as it otherwise starts right away with the first level:

https://github.com/kaplayjs/kaplay/pull/1005/changes#diff-cc6410035824d56b4a435bc960e1521432538f519440a5dfada744469c861a21R483-R485

```js
onLoad(() => {
    go("win", currentIdx);
})
```

examples/sokoban.js
```diff
@@ -7,75 +7,482 @@
...
+            size: 50,
+            align: "center",
+        }),
+    ]);
+
+    add([
+        anchor("center"),
+        pos(center().sub(0, 160)),
+        text(
+            `You successfully re-stocked\n the [o]soks[/o] in [o]${moves}[/o] moves!`,
```
imaginarny: You used `moves` variable that tracks the moves count of the current level that gets reset at the start of each level which means this number here is the moves count from the last level only. 

So you need a separate global variable that counts the moves total and add moves value to it e.g. `movesTotal += moves;` before you switch the scene - probably among the lines ~[238](https://github.com/kaplayjs/kaplay/pull/1005/changes#diff-cc6410035824d56b4a435bc960e1521432538f519440a5dfada744469c861a21R238)+ and use that variable here instead.

lajbel:
Thank you super cool

examples/sokoban.js
```diff
@@ -7,75 +7,575 @@
...
+                        color: Color.fromHex("#ea6262"),
+                    },
+                },
+            },
+        ),
+    ]);
+
+    // The fewest moves possible to beat the whole game
+    if (movesTotal <= 45) {
+        c;
```
imaginarny, line 443: RIP

amyspark-ng:
i trust your word on it

imaginarny:
Have you noticed the `c;`?

amyspark-ng:
is the fix pushed i can't see it

PR #1010, opened 2026-01-13
feat: add calculate() method to fps counters so advanced users can make a fps graph

dragoncoder047:
idk if this deserves a changelog cause "for advanced users only" ?


lajbel:
It deserves a changelog

PR #1011, opened 2026-01-14
feat: improve tex packer using used-rects

dragoncoder047:
should (?) support removing textures and then filling the gaps (unimplemented)

## Summary

- [X] Changeloged


dragoncoder047:
For reference this is how it does:

platformer example before:

Screenshot 2026-01-13 at 9 20 02 PM

platformer example after:

Screenshot 2026-01-13 at 9 21 04 PM

PR #1036, opened 2026-02-07
feat: allow slice9'ed drawSprite

dragoncoder047:
allow slice9'ed sprites in drawSprite()

## Summary

- [X] Changeloged


src/gfx/draw/drawSprite.ts
```diff
@@ -60,20 +64,105 @@ export function drawSprite(opt: DrawSpriteOpt) {
     }
 
     // TODO: slow
-    const spr = resolveSprite(opt.sprite);
+    var spriteData: SpriteData;
```
mflerackers: Should be let

mflerackers (changes requested):

lajbel:
Wouldn't be better to create a separated function like `draw9SliceSprite()`?

dragoncoder047:
> Wouldn't be better to create a separated function like `draw9SliceSprite()`?

possibly, but then it would make sense to also have one for drawTiledSprite(), etc., idk if it's worth it.

lajbel:
I think it does. We reduce overhead in sprite drawing functions, what do you think @mflerackers?

dragoncoder047:
i think it conflicts with the way sprites are defined, that is, the slice9 parameters are stored in the sprite metadata, so what would happen if you called draw9SliceSprite with a sprite with no slice9 metadata. Or vice versa

mflerackers (approved):

PR #1043, opened 2026-02-19
fix: account for tilePos.x getting out of the level boundary

benhuangbmj:
`levelComp.getAt` returns incorrect results when `tilePos.x` is beyond the scope of the level. 
## Summary
  When the `tilePos` is beyond the scope of the level, the `getAt` method should return an empty array. The `tile2Hash` function already accounts for `tilePos.y` getting outside of the level. However, it doesn't work for `tilePos.x` properly. I added a simple check in the `getAt` method without touching any other part of the code to avoid unexpected breaking behavior.
- [x] Changeloged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,8 @@ So your change should look like
   `drawSprite` (#1026) - @benhuangbmj
 - Fixed `onClick()` and `onCollide()` tag variants no longer working -
   @mflerackers
+- Fixed `levelComp.getAt` to account for `tilePos.x` getting outside of the
+  level boundary - @benhuangbmj
 
 ## [4000.0.0-alpha.26] - 2026-01-12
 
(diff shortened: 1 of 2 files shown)


dragoncoder047:
> The tile2hash function already accounts for tilePos.y getting outside of the level. 

you sure about that?

https://github.com/benhuangbmj/kaplay/blob/7dfe1a634d0d821459ffcf136a6f48699a57d5c7/src/ecs/components/level/level.ts#L141

benhuangbmj:
> > The tile2hash function already accounts for tilePos.y getting outside of the level.
> 
> you sure about that?
> 
> https://github.com/benhuangbmj/kaplay/blob/7dfe1a634d0d821459ffcf136a6f48699a57d5c7/src/ecs/components/level/level.ts#L141

What I meant was, suppose the level has _m_ rows and _n_ columns, then its tiles are indexed from 0 to _m*n_  - 1. As long as  0 <= `tilePos.x` < _m*n_, then whenever `tilePos.y` >= _m_ or `tilePos.y` < 0, `tile2Hash` will return a number that is beyond the proper index range, henceforth `getAt` will just return an empty array.

However, without the "0 <= `tilePos.x` < _m*n_" check, `tile2Hash` may potentially return an index within the proper index range even if `tilePos` itself is beyond the level boundary. And that's why I proposed the current fix.

lajbel:
Would you give an example of the bug?

PR #1062, opened 2026-03-25
feat: add back no repack option, also minor speedup of repacker

dragoncoder047:
- [X] Changeloged

### DIFF
--- a/src/assets/sprite.ts
+++ b/src/assets/sprite.ts
@@ -94,11 +94,13 @@ export interface LoadSpriteOpt {
      */
     filter?: TexFilter;
     /**
-     * If you've already packed your spritesheet, you can set this to true to tell KAPLAY not to repack it
+     * If you've already packed your spritesheet, you can set this to false to tell KAPLAY not to repack it
      * when loading it into the main texture. However, this may negatively impact rendering performance if
      * it causes future sprites to not fit and be moved to a different GPU texture.
      *
      * This doesn't apply if you use singular: true, because the entire image will be its own texture then.
+     *
+     * @default true
      */
     repack?: boolean;
 }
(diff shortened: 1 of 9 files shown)


Stanko:
I tested `repack: false` on my game and it works as expected 👍

src/assets/sprite.ts
```diff
@@ -93,6 +93,14 @@ export interface LoadSpriteOpt {
...
      */
     filter?: TexFilter;
+    /**
+     * If you've already packed your spritesheet, you can set this to true to tell KAPLAY not to repack it
+     * when loading it into the main texture. However, this may negatively impact rendering performance if
+     * it causes future sprites to not fit and be moved to a different GPU texture.
+     *
+     * This doesn't apply if you use singular: true, because the entire image will be its own texture then.
+     */
+    repack?: boolean;
```
imaginarny, line 105: This is confusing. I'm setting `repack` to `true` to not to repack it? Shouldn't it be the other way around, "true, (use) repack"? Not logic like "true, I repack**ed** it already myself, no need". In that case, maybe `repacked` would be more clear but I'm still not sure I understand it to say surely. Or to be clear you are preventing the sprite from getting repacked. Cause currently it's as if used as "irrepackable: true" noun. Would like input of others on this.

So when @Stanko above tested his game with repack set to true, he tested with repack disabled? Was there then any speed improvement for you with our repack? I thought there was until I looked into that code comment.

The other thing to think about is, should even spritesheets and sprite atlases be repacked? This caused issues, right? Since you define slices or positions you expect the frames to be at. How is it then safely repacked? Instead of it being placed as is in the texture as one would expect. Repack idea sounds neat on paper, but applied seems like stuff like that is already too late and heavy to happen in runtime and not "build" time, thinking about whole repack situation more and more. So it seems that next time we will just keep recommending ppl to set this option to false/true or whatever it ends up on their spritesheets?

dragoncoder047: whoopsie, I originally wrote the doc comment when it was called `dontRepack`

Stanko: I updated my comment to reflect new changes to avoid further confusion

lajbel (approved):

PR #1064, opened 2026-03-26
feat: text character check index

dragoncoder047:
- [X] Changeloged


dragoncoder047:
@imaginarny would appreciate your help on resolving merge conflicts here since you rewrote the text component

imaginarny:
Besides merging and conflict resolving I did a few changes as well:

- Since we everywhere abbreviate character to char, I changed the api naming and made it easier to use
	- Renamed `pointToCharacter()` to `pointToCharIndex()` since it returns index
	- Added `pointToChar()` as well, that returns char data right away (FormattedChar), so you don't have to do `obj.formattedText().chars[index]` manually (but still can with `pointToCharIndex()`)
- Changed the chars inspect bounding boxes drawn
	- Refactored it to use picture api to go from 10 FPS to all fps (for ~A4 page amount of text)
	- Changed the yellow color (couldn't be seen on white and otherwise a little bit too overwhelming) to char color itself, lerped slightly to gray, fixed lightness, and with lowered opacity instead, so it's way more subtle
	- Changed that it's only drawn if text includes tags, uses transform or styles
	- But I'm still on a fence if it isn't too much and if it shouldn't be only enabled either as a text or global opt, with automatic option like that or some different logic, or idk..
- Updated `textClick` example to be touch compatible

PR #1077, opened 2026-04-18
fix: broadphase objects with stay cleared on scene switch

imaginarny:
Turns out that onSceneLeave has to be removed completely from broadphase and its objects can't be cleared there since there are also objects with stay(). Addition to #1074.

- [x] Changeloged


mflerackers (approved):
Did you try if a stay object is clickable after scene switch?

imaginarny:
Ye, added new playtest based on the previous, those beans are now added in the scene only once with `stay()` instead and all are clicklable, on scene switch as well.

PR #1078, opened 2026-04-20
fix: tileMove-based functions on a level tilemap

### LINKED ISSUE #903: bug: calling tileMove-based functions on a level tilemap causes an error
### Bug Description

Calling functions based on `tileMove` causes an error if `level.getAt()` is not called first. 

Some preliminary investigation seems to suggest that the private `spatialMap` is not being created during the creation of the `level` object in `addLevel`.

This bug also appears to be present in v. 3001 as well.

### Version

4000.0.0-alpha.21

### Playground Link

https://play.kaplayjs.com/?code=eJyNUktPAjEQvvdXTOqBlmxYMIYDBOPjYiImRk04GA51d1waSrtpC4QY%2FrvtdhXQg86hmdf3dV55Ds9eWO9AgMYtVGKFZClqJXaMjwnJc5gaUYJwDr0jKug3KHSMRf25ttIjo84jKpoBzQuL27wxe7Wu6I%2B8ygaiQ15jfuXFvx7EEkMlCjeoSGG080mHCYiynEaVvRIIQs%2Bi0Kw1wvMfo8XMM%2FhoPF4qnMnSL0YwvMi%2BXXcoq4U%2F%2BLZSlYWw5UuIjYBxmFxCKiOKO2mOJ8j8wOZG7W%2BpBvqb4ZgljbJlOWJqwFd%2FgN%2FCbk6wjbYne07accbNog3zbObac7XYahZ4M9hgcc4GGQw4J3m3Ac6kX5i1B7%2BQDpTUmIHQgNYaC6Yo1taB0bDEXW3RhTbTlG%2BF1saDxXA3tTU1Wi8xZL6DXisFLAakrqAz7HBKujlJpVTorz1rquhn0OfxJIy%2Bx91jJGe0DDWm3tM0Uye9ldngU1xXvMk9H38CQDC%2Fxg%3D%3D&version=4000.0.0-alpha.21

### Extra information

_No response_

### Summary

- [ ] Fixed in v4000
- [ ] Fixed in v3001

SobennaStory:
Closes #903 

Added options to create spatial map and tilepath-based maps to levelcompopt, as well as more descriptive errors in spatialmap functions. Users should be able to more easily understand and proactively handle tilemove and spatialmap based issus

### DIFF
--- a/src/ecs/components/level/level.ts
+++ b/src/ecs/components/level/level.ts
@@ -112,6 +112,14 @@ export interface LevelCompOpt {
         [sym: string]: (pos: Vec2) => CompList<Comp>;
     };
     /**
+     * Enable spatial map after adding level tiles
+     */
+    spatialMap?: boolean;
+    /**
+     * Enable cost, connectivity, and edge maps after adding level tiles
+     */
+    tilePathMaps?: boolean;
+    /**
      * Called when encountered a symbol not defined in "tiles".
      */
     wildcardTile?: (
@@ -149,7 +157,16 @@ export function level(map: string[], opt: LevelCompOpt): LevelComp {
         }
     };
 
+    const ensureSpatialMapExists = () => {
+        if (!spatialMap) {
+            throw new Error(
+                "Spatial map is not initialized. Call level.getSpatialMap() before using spatial map functions, or enable the 'spatialMap' option in level().",
+            );
+        }
+    };
+
     const insertIntoSpatialMap = (obj: GameObj) => {
+        ensureSpatialMapExists();
         const i = tile2Hash(obj.tilePos);
         if (spatialMap![i]) {
             spatialMap![i].push(obj);
@@ -160,6 +177,7 @@ export function level(map: string[], opt: LevelCompOpt): LevelComp {
     };
 
     const removeFromSpatialMap = (obj: GameObj) => {
+        ensureSpatialMapExists();
         const i = tile2Hash(obj.tilePos);
         if (spatialMap![i]) {
             const index = spatialMap![i].indexOf(obj);
@@ -351,6 +369,16 @@ export function level(map: string[], opt: LevelCompOpt): LevelComp {
                     this.spawn(key, vec2(j, i));
                 });
             });
+
+            if (opt.spatialMap) {
+                createSpatialMap(this);
+            }
+
+            if (opt.tilePathMaps) {
+                createCostMap(this);
+                createEdgeMap(this);
+                createConnectivityMap(this);
+            }
         },
 
         tileWidth() {
@@ -737,9 +765,9 @@ export function levelFactory(data: any) {
         const d = data.tiles[key];
         const tags = d.tags;
         opt.tiles[key] = (pos: Vec2) => {
-            const comps: Comp[] = Object.keys(d).filter(k => k != "tags").map(
-                id => deserializeComp(id, d[id]),
-            );
+            const comps: Comp[] = Object.keys(d)
+                .filter((k) => k != "tags")
+                .map((id) => deserializeComp(id, d[id]));
             return [...comps, ...tags];
         };
     }
--- a/src/ecs/components/level/level.ts
+++ b/src/ecs/components/level/level.ts
@@ -1,17 +1,17 @@
 import type { KEventController } from "../../../events/events";
 import {
-    raycastGrid,
-    type RaycastResult,
-    vec2,
-    type Vec2Args,
+  raycastGrid,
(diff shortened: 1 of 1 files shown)


src/ecs/components/level/level.ts
```diff
@@ -111,6 +111,14 @@ export interface LevelCompOpt {
...
         [sym: string]: (pos: Vec2) => CompList<Comp>;
     };
+    /**
+     * Enable spatial map after adding level tiles
+     */
+    spatialMap?: boolean;
+    /**
+     * Enable cost, connectivity, and edge maps after adding level tiles
+     */
+    tilePathMaps?: boolean;
```
dragoncoder047:
```suggestion
```

src/ecs/components/level/level.ts
```diff
@@ -351,6 +369,16 @@ export function level(map: string[], opt: LevelCompOpt): LevelComp {
...
+
+            if (opt.spatialMap) {
+                createSpatialMap(this);
+            }
+
+            if (opt.tilePathMaps) {
+                createCostMap(this);
+                createEdgeMap(this);
+                createConnectivityMap(this);
+            }
```
dragoncoder047:
```suggestion
```

src/ecs/components/level/level.ts
```diff
@@ -149,7 +157,16 @@ export function level(map: string[], opt: LevelCompOpt): LevelComp {
...
     };
 
+    const ensureSpatialMapExists = () => {
+        if (!spatialMap) {
+            throw new Error(
+                "Spatial map is not initialized. Call level.getSpatialMap() before using spatial map functions, or enable the 'spatialMap' option in level().",
+            );
+        }
+    };
+
```
dragoncoder047:
```suggestion
    const ensureSpatialMapExists = () => {
        if (!spatialMap) {
            createSpatialMap(this);
        }
    };

```

If the spatial map doesn't exist yet ...why not just create it now?

src/ecs/components/level/level.ts
```diff
@@ -737,9 +765,9 @@ export function levelFactory(data: any) {
         const d = data.tiles[key];
         const tags = d.tags;
         opt.tiles[key] = (pos: Vec2) => {
-            const comps: Comp[] = Object.keys(d).filter(k => k != "tags").map(
-                id => deserializeComp(id, d[id]),
-            );
+            const comps: Comp[] = Object.keys(d)
+                .filter((k) => k != "tags")
+                .map((id) => deserializeComp(id, d[id]));
```
dragoncoder047: Since this is just formatting changes, please revert this. We use dprint for formatting, not whatever your editor wants. Open a terminal and run `pnpm fmt`.

dragoncoder047 (changes requested):

src/ecs/components/level/level.ts
SobennaStory: The this inside an arrow function is lexically bound, so it won't refer to the level instance here. 
The throw approach sidesteps the problem by making it the caller's responsibility to initialize the map first.

I also mused on 
1. Simply creating the spatial map when the component is attached via add() 
2. Making the tile.tilemove functions throw errors if no spatial map is found in their responsible level
3. Creating a spatial map whenever level.spawn() is called.

What do you think the best approach is? To implement the above suggestion we'd need to change the props of insertIntoSpatialMap and removeFromSpatialMap which may break other usages of the functions?

dragoncoder047: shoot, forgot about that part, that's what I get for not checking this out locally.

to your muses

1. not necessary for levels that don't use it
2. please no errors when it could just be created
3. not necessary for levels that don't use it

I mostly just wanted to reduce the number of edits, so just every place that `ensureSpatialMapExists()` is called, just replace it with `if (!spatialMap) createSpatialMap(this)` or `if (!spatialMap) createSpatialMap(obj)`, you might have to make wrappers for `insertIntoSpacialMap` and `removeFromSpatialMap`, instead of just using the functions directly, make them wrappers to get the `this` -
```diff
-   insertIntoSpatialMap,
+   insertIntoSpatialMap(obj) { insertIntoSpatialMap(obj, this); }
```

SobennaStory:
figuring out fmt

src/ecs/components/level/level.ts
SobennaStory: resolved

amyspark-ng:
@SobennaStory Marked as waiting for author as it looks like there's unresolved conversations, if i'im the wrong please tell me so i can mark ready for review and re-request a review

Issue #326, opened 2024-08-05
bug: mouseDeltaPos() isn't reset to (0, 0) on frames that have no mousemove event


dragoncoder047:
put in kaplayground:

```js
kaplay();
const cursor = add([
    circle(10),
    color(RED),
    pos(500, 500)
]);
onUpdate(() => {
    cursor.move(mouseDeltaPos().scale(1 / dt()));
});
```

then jerk your mouse (might work better if you have a trackpad). The red ball keeps moving even though the mouse stopped moving!

PR #1079, opened 2026-04-21
feat: added scrollDelta() function to app state

### LINKED ISSUE #981: feat: scrollDelta()

we have mouseDeltaPos() so you can do obj.moveBy(mouseDeltaPos()) to make it drag with the mouse. For scrollviews it would also be nice to have obj.moveBy(scrollDelta()) to enable this instead of just having to catch it via an onScroll handler

### Any more information?

_No response_

SobennaStory:
closes #981 

Added scroll delta as a pollable input, consistent with how mouse delta position works. Instead of only being able to react to scroll via the onScroll callback, you can now also read it directly like any other input state.

### DIFF
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -205,6 +205,7 @@ export const initAppState = (opt: {
         capsOn: false,
         mousePos: new Vec2(0),
         mouseDeltaPos: new Vec2(0),
+        scrollDelta: new Vec2(0),
         keyState: new ButtonState<Key>(
             "keyPress",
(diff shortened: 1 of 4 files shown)


dragoncoder047 (changes requested):
1. You don't need to create a new Vec2 every time we get a scroll, you can just `.set()` it to the new values.

~2. You need to set `scrollDelta` to zero on frames where we don't have a `wheel` event. I don't want another #326.~ My dumb ass can't read. Never mind.

Issue #1075, opened 2026-04-17
bug: overall sprite atlas padding is missing


imaginarny:
### Bug Description

Since the new tex packer, texture atlas is missing the overall padding, there are only gaps between the sprites.
Meaning, sprites in the texture at the edge are only padded on the rest of the sides.

That can cause a repeating px texture on drawn frames with an offset (not only intentionally, but also due to rounding inaccuracies as well), also the reason `spriteAtlasPadding` kaplay option was added.

You can switch to the previous version in KAPLAYGROUND to see the difference.

### Version

4000.0.0-alpha.27

### Playground Link

[...]

PR #1089, opened 2026-05-03
fix: refresh packer for direct referenced sprites resolving

dragoncoder047:
also some trivial optimizations

this is so trivial it probably doesn't even need changelog since it was fine in a27 so the bug will be fixed by the next release if this is merged before then

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,3 +1,7 @@
+allowBuilds:
+  dprint: true
+  esbuild: true
+  puppeteer: true
 overrides:
   axios@>=1.0.0 <1.15.0: '>=1.15.0'
   basic-ftp@<=5.2.1: '>=5.2.2'
--- a/src/assets/sprite.ts
+++ b/src/assets/sprite.ts
@@ -243,6 +243,9 @@ export function resolveSprite(
     //     return Asset.loaded(src);
     // }
     else if (src instanceof Asset) {
+        // might be sometimes unnecessary here but if we're getting the sprite data directly
+        // then obviously we might be wanting to draw it.
+        _k.k.onLoad(() => _k.assets.packer.syncIfPending());
         return src;
     }
     else {
--- a/src/gfx/TexPacker.ts
+++ b/src/gfx/TexPacker.ts
@@ -264,7 +264,7 @@ export class TexPacker {
         packer._used.set(-1, {
             rect: new Rect(new Vec2(width - 1, height - 1), 1, 1),
             tex: tex,
-            used: 0,
+            used: UsedCorner.BOTTOMLEFT | UsedCorner.TOPRIGHT, // it's bottom right so nothing can go right or below it
         });
         return {
             tex,
--- a/src/gfx/formatText.ts
+++ b/src/gfx/formatText.ts
@@ -230,6 +230,7 @@ function updateFontAtlas(font: FontData | string, ch: string) {
         const img = c2d.getImageData(0, 0, w, h);
 
         atlas.font.map[ch] = _k.assets.packer.add(img, atlas.font.filter);
+        _k.assets.packer.syncIfPending();
 
         atlas.maxHeight = Math.max(atlas.maxHeight, h);
     }
@@ -471,8 +472,6 @@ export function formatText(opt: DrawTextOpt): FormattedText {
         }
         th += thisLineHeight;
     }
-
-    _k.assets.packer.syncIfPending();
 
     return {
         width: tw,
--- a/tests/auto/texturesync.spec.ts
+++ b/tests/auto/texturesync.spec.ts
@@ -1,0 +1,33 @@
+import { beforeAll, describe, expect, test } from "vitest";
+
+describe("Components validation in add()", async () => {
+    beforeAll(async () => {
+        await page.addScriptTag({ path: "dist/kaplay.js" });
+    });
+
+    test(
+        "direct sprite referencing should force a texture flush",
+        async () => {
+            async function fn() {
+                return page.evaluate(() => {
+                    const k = kaplay();
+                    return new Promise(resolve => {
+                        // hacky since we can't use spyOn mocking
+                        var called = false;
+                        k._k.assets.packer.syncIfPending = function() {
+                            called = true;
+                        };
+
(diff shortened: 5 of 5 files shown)


imaginarny:
Ok so, is this then a follow up on recent fixes that got merged in master as intended a27.1 while master currently contains only fixes for a27? Since you say its fine in a27 and imply it isn't in master anymore. If so, should be merged before that as well.

Second question, how do I test it to confirm fix and what would be the real use case for directly referencing sprites to see? (like in #1075?)

dragoncoder047:
this would be things like where you would use

```js
const bean = loadBean();

add([
    sprite(bean);
]);
```

instead of 

```js
loadBean();

add([
    sprite("bean");
]);
```

lajbel:
@dragoncoder047 I think this use case could use an automated test?

dragoncoder047:
how the heck would you automate a test like that? do you mean another non-automated playtest?

lajbel:
Use vitest. Anyway I don't know if it will work, just saying as we're pointing to have more stuff tested

dragoncoder047:
it randomly fails random files, but i think i got something that works

imaginarny:
Well, I realized this also changes behavior of when sprites are added to the texture. Initially I thought it was like it is now, but it wasn't (unlike fonts) and I was somewhat relieved then.

So, before a sprite was added as soon as you did loadSprite. Now it is added only if you really draw it. So the question is, should there be another perf concern and more testing done for that? Like is tex packer fast enough when sprites are added on fly? Like a whole spritesheet too. You already mentioned that for the text there is a frame drop since letters worked like that. Would like to see a playtest with measures when a lot is loaded and then sprites are drawn after some delay, one by one or multiple at once, some heavier ones as well as dungeon spriteAtlas we have. What will performance then look fps-wise.

dragoncoder047:
No, it loads the sprite data on to the `canvas` as they're loaded. The thing I was optimizing was cutting how often it synced that canvas to the GPU, which was really, really slow compared to adding another image to the canvas.

imaginarny:
Alright, so they actually are in the texture, but then my saved ref spriteData is outdated now. Heh, this is something similar to what MF said before (for changes in frames/quad, which don't happen), but tests showed it wasn't needed. 

So when you reference spriteData directly and e.g. use its texture, it will get outdated after you load more stuff. That's where I saw it wasn't updating and thought new sprites weren't added to the texture. After calling getSprite again, I see they are there - if you say repack already happened on loadSprite and not on my second getSprite call itself. So if that's the case, there isn't real life usage for something like that, since the sprite itself, its frame/quad doesn't change anyway, and this is where optimization is coming from?

dragoncoder047:
> but then my saved ref spriteData is outdated now.

No, it won't be. This PR doesn't change anything about repacking!! The texture packer never automatically moves or deletes things after they've been packed. It just defers syncing the texture data to the GPU until it knows you might want to draw the sprite (i.e. calling getSprite directly, and this worked for "named" sprites since the sprite component calls getSprite when you give it a string, but I forgot about directly referencing sprite data, which addKaboom does).

imaginarny:
It won't be? I didn't say it's repacking itself, I even said why it made me think in the first place and then question what's happening. So idk if I wasn't clear, but why, when I do this, Mark isn't drawn:

```js
loadBean();

onLoad(() => {
    let bean = getSprite("bean");

    loadSprite("mark", "/sprites/mark.png");

    onDraw(() => {
        drawUVQuad({
            tex: bean.tex || bean.data.tex,
            width: 2000,
            height: 2000,
        });
    });
});
```
He is there in a27, and now another interesting part, he would be there on master too if there was some text component added as well (with the nearest filter obviously), BUT not on this branch, that is another difference.

So I have to call it again for mark to appear, e.g.:
```js
wait(3, () => bean = getSprite("bean"));
```

Which I asked if it's the resulting optimization in question.

[play.kaplayjs.com/?code=eJx1T00Lgk...&version=master](https://play.kaplayjs.com/?code=eJx1T00LgkAQvfsrhj2tICl1M%2BwQHbtE1H1iF120VXTCIvvvza5YGPQuO%2FM%2BmLclNhU%2BZLgOgqpGtdVo%2FVLbPa9ShpBt4BkAo9IEF9Yhg1zTsWkNaSkcI1zCWzgzCVdsSxGBiDtPdLEjFo3NP%2B7a7lrsZzccFJOn8%2BHG57%2BkA%2Bl76hsseIJhGGeFhI6IZt7eKCpSWCZJMhcKbfKCfpUXd5peP%2FRoSK4iGMv9%2Bza7XeINo4dU5w%3D%3D&version=master)

dragoncoder047:
Well, for something like what you're doing (drawing the ENTIRE texture) it doesn't work, but who does that. If you wanted to use mark, you'd getSprite("mark") in order to have the quad that you can pass to particles() etc.

imaginarny:
So, this is what I summed up in my reply. So your reply could actually be Yes, as you said. Instead "No, it won't be.". :)

dragoncoder047:
It's not outdated if you mean it surreptitiously references the wrong quad on the texture, that never changes. The texture itself may be updated later but never in a place that an existing sprite data references.

imaginarny (approved):

imaginarny:
Right, so this is done.

Also, MF has another packer optimization idea that you might be interested in, since we are already discussing here, that could or not end up in a27.1 as well, if you are up for it. Currently, it's the whole texture that is sent on sync to the gpu and since `texSubImage2D` is used, you could keep track of the union of the rects and sync only the sub image instead and save a lot of bandwidth. E.g., for a 32x32 sprite you are now sending 16.777.216 bytes instead of 4096 bytes.

dragoncoder047:
Hmm, I just looked at that, and because we're sending an `ImageSource` to `texSubImage2D()` we can't use the form with `width` and `height`, since that only works with a typedarray and not an `` or `<canvas>`. Perhaps there is another way.

imaginarny:
[`createImageBitmap`](https://developer.mozilla.org/en-US/docs/Web/API/Window/createImageBitmap), do you want an issue or discuss on DC if needed or just do PR/draft when you got something? Since I'm just merging this, ty. 👍

PR #1096, opened 2026-05-20
fix: mouse coordinates were off when canvas wasn't rendered at it's natural size

### LINKED ISSUE #1001: bug: when Kaplay's canvas is resized, mouse event coordinates are off
### Bug Description

As I said on discord, I'm not sure if this is something you plan to change, but here is my issue.

I use CSS to resize the game's canvas, in order to fill the entire width of the screen. Because the size now (704x396) is different than canvas' natural size  (960x540) mouse coordinates are off.

Here you can see in the screenshot, game is showing the overlay for the big boss at the top and I'm hovering right of it

Image

I think coordinates should be remapped, something like this:

```ts
  const rect = k.canvas.getBoundingClientRect();
  const scale = k.width() / rect.width;

  const x = (e.clientX - rect.left) * scale;
  const y = (e.clientY - rect.top) * scale;
```

I understand this could be mitigated by resizing the game on window resize, but that introduces a different set of problems.

### Version

4000.0.0-alpha.25

### Playground Link

_No response_

### Extra information

_No response_

### Summary

- [ ] Fixed in v4000
- [ ] Fixed in v3001

Stanko:
Closes #1001

Update viewport coordinate conversion

- Cache canvas display scale in app state and refresh it on resize
- Use cached scale in `canvasToViewport` and `viewportToCanvas`
- Make `viewportToCanvas` the inverse of `canvasToViewport`
- Add `viewportToCanvasLocal` for debug overlay viewport-local rendering
- Preserve letterbox offset handling in both conversion paths
- [x] Changeloged


imaginarny (approved):
This is a great help! I meant to look into this when someone else recently asked about centering canvas (sry, I think I missed when you asked about custom CSS sizing) and it was possible but only with `letterbox` set to true, even though it didn't make much sense to use since the canvas dimensions were fixed. But, it was crucial as a workaround for keeping the reported coords in responsive correct.

Will paste the dirty trick here as well so it's easier to follow xD.

```js
kaplay({
    width: 800,
    height: 600,
    letterbox: true,
});

canvas.parentElement.style.display = "flex";
canvas.style.cssText += `
    margin: auto;
    width: min(${width()}px, 100%, calc(100svh * ${width()} / ${height()}));
    height: min(${height()}px, 100%, calc(100svw * ${height()} / ${width()}));
`;
```
It behaves exactly like letterbox when window scaled on either axis, but without canvas-visible letterbox. Now, `letterbox` opt isn't required anymore. Although that trick itself does not matter that much as the overall issue.

But the reason I'm writing this is that you also made it easier to pin-point another bug I've found during that workaround in fullscreen vs touch, so I might ping you soon once I fix it, for some additional validation. 🙂

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ So your change should look like:
 
 - Fixed `TimerController.timeLeft` returning elapsed time instead of remaining
   time (#1082) - @nojaf
+- Fixed mouse coordinates not being calculated properly when canvas is resized by CSS and wasn't rendered at it's natural size (#1096) - @Stanko
```
imaginarny:
```suggestion
- Fixed mouse coordinates not being calculated properly when canvas is resized by CSS and wasn't rendered at its natural size (#1096) - @Stanko
```
Oops, I've found another typo, `its` possessive form should be there.

imaginarny (changes requested):

Stanko:
Glad I could help :) I use a very similar CSS hack with my game (relying on `object-fit: contain`) and that's how noticed that mouse coordinates don't line up.

imaginarny (approved):

PR #1101, opened 2026-05-31
fix: button release on blur

### LINKED ISSUE #1094: bug: `isKeyDown` and `isButtonDown` get stuck when cmd-tabbing on mac
### Bug Description

When holding a key or a button and use cmd-tab to change to other application, `isKeyDown()` (or `isButtonDown()`) get stuck and report as`true` until the key/button is pressed again. I created a small example at playground where you can test it.

I'm not sure if this is a mac only issue.

### Version

4000.0.0-alpha.27

### Playground Link

https://play.kaplayjs.com/?code=eJydkU9rwjAYh%2B%2F5FC%2F1YAK1tiA7KM6LN4%2By09ghS2ItcUlIU7QMv%2FvyR6EIRVxuCb%2Fn9yRv5nPYO2pdCxSUOENNfwRiWrUOJKxBUnOiPf5F4Nc3ZbK2ulN8CdNJFdc0R1eyQjfkJA4uUAXlHH%2FKwugWy4IJ5YTFJJ7OqkWZQ0lIHmKKHbXFWUpk8UwbyhrX47JYxL0TF4ez0JyRL69KJtvUx2eq%2F5li81DVmSceb5m9vezpzFDC9VmNa16sDmWpHPmI%2BjCcOoExgfU7pK8M47zDUdu0O9FvPXcfNWyggiX44lUk4ljGkNvMHpnOjAHh8Y%2FpcOuxfHrRkLgShNAfZ3LQmw%3D%3D&version=4000.0.0-alpha.27

### Extra information

_No response_

### Summary

- [ ] Fixed in v4000
- [ ] Fixed in v3001

Stanko:
Relates to #1094 

I would like you folks to take a look before I convert it to a "real" pull request.

I added "release all" buttons for keyboard and mouse. These events can "drop" when focus leaves the canvas. On the other hand gamepad is polling for events so I made sure to skip those (if there are any).

### DIFF
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -89,6 +89,11 @@ export class ButtonState<T = string, A = never> {
             state.events.trigger(this._releaseEv as any, btn, this._arg);
         }
     }
+    releaseAll(state: AppState) {
+        for (const btn of this.down) {
+            this.release(btn, state);
+        }
+    }
 }
 
 class GamepadState {
@@ -978,7 +983,38 @@ export const initApp = (
     const docEvents: EventList<DocumentEventMap> = {};
     const winEvents: EventList<WindowEventMap> = {};
 
+    let releaseHeldInputsQueued = false;
+
+    function releaseHeldInputs() {
+        state.buttonHandler.releaseKeyboardMouse(state);
+        state.keyState.releaseAll(state);
+        state.mouseState.releaseAll(state);
+    }
+
+    function queueReleaseHeldInputs() {
+        if (releaseHeldInputsQueued) {
+            return;
+        }
+        releaseHeldInputsQueued = true;
+        state.events.onOnce("input", () => {
+            releaseHeldInputsQueued = false;
+            releaseHeldInputs();
+        });
+    }
+
+    function releaseHeldInputsOnFocusLoss() {
+        // Release all inputs immediately when blur/hide happens, 
+        // and then queue it to to catch any queued events on the next input tick,
+        // that wouldn't be processed otherwise
+        releaseHeldInputs();
+        queueReleaseHeldInputs();
+    }
+
     const pd = opt.pixelDensity || 1;
+
+    canvasEvents.blur = () => {
+        releaseHeldInputsOnFocusLoss();
+    };
 
     canvasEvents.mousemove = (e) => {
         // 🍝 Here we depend of GFX Context even if initGfx needs initApp for being used
@@ -1259,9 +1295,14 @@ export const initApp = (
             state.events.trigger("show");
         }
         else {
+            releaseHeldInputsOnFocusLoss();
             state.isHidden = true;
             state.events.trigger("hide");
         }
+    };
+
+    winEvents.blur = () => {
+        releaseHeldInputsOnFocusLoss();
     };
 
     winEvents.gamepadconnected = (e) => {
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -1003,7 +1003,7 @@ export const initApp = (
     }
 
     function releaseHeldInputsOnFocusLoss() {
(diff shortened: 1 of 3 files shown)


src/app/app.ts
```diff
@@ -89,6 +89,11 @@ export class ButtonState<T = string, A = never> {
             state.events.trigger(this._releaseEv as any, btn, this._arg);
         }
     }
+    releaseAll(state: AppState) {
+        for (const btn of [...this.down]) {
```
imaginarny:
```suggestion
        for (const btn of this.down) {
```

No need to spread Set in an intermediate array here, it's iterable.

src/app/app.ts
```diff
@@ -978,8 +983,36 @@ export const initApp = (
...
+        releaseHeldInputsQueued = true;
+        state.events.onOnce("input", () => {
+            releaseHeldInputsQueued = false;
+            releaseHeldInputs();
+        });
+    }
+
+    function releaseHeldInputsOnFocusLoss() {
+        releaseHeldInputs();
+        queueReleaseHeldInputs();
```
imaginarny, line 1010: What is the reason `releaseHeldInputs()` is called twice? Once right away then again in input event queue. Wouldn't it be enough with the queued one? 🤔

imaginarny (changes requested):
This was on todo so it would be great help again as well. 🙂

Will test some more once its updated and could merge it before #1106 so you have less work merging conflicts. Plus, I have one another in progress that could cause troubles as well, so would make sense to merge this one first.

src/app/app.ts
Stanko: True, it is a force of habit of avoiding mutations 🙈

src/app/app.ts
Stanko: That's why I wanted you to take a look 😅 but from what I gathered keeping only the queue would leave held inputs stale until the next input tick. And in the case of  `visibilitychange` to hidden that tick may not run until the tab is visible again. So the synchronous one triggers before the `hide` event and makes sure it's event handlers don't see stale state.

I added a comment in the code to try to explain it better. 

But I would love your input on it.

imaginarny: There are 2 scenarios happening:

1. Input was not held down yet, it was pressed at the same time when blur/hide happens
    - When you call release immediately, the input events weren't processed yet so you try to release before they are registered and then they won't release and stay pressed. This can happen in the same frame. So here it's not like the queued one catches the rest, the first one never releases any.
2. Input was held down already when blur/hide happens
    - When you call release immediately, it releases "mid" frame. If it was some different use case, it could cause issues, that's why all inputs are processed in input event I think, like e.g. you could trigger input rightfully and clear it before some other api would detect it was pressed. 🤔 But doesn't matter much here anyway, I wouldn't call it stale state since by design it clears in the beginning of the frame, but would have to verify the timing of the rest of input apis.

So it runs 4x on hide, 3x on win blur, 2x on canvas blur only. Hide and complete win blur don't matter and even if you used html UI outside of canvas, one intermediate array that gathers all inputs is negligible. Also no need to distinguish between those scenarios, not even reliably possible/would be heavier to. Just to include some insight. I think in this case it's fine then. Would like @lajbel to take a look if there was something overlooked.

Stanko: Sounds good, but let's wait for lajbel to chime in.

lajbel: LGTM 👍🏻 

OK to proceed @Stanko

imaginarny:
Would merge once you update the branch and changelog it.

Stanko:
Done and done 🙌

imaginarny (approved):
Thank you. 👍

PR #1102, opened 2026-06-01
feat: improve tex packer speed by like 3x

dragoncoder047:
speeds up the tex packer by like >=3x i think

on textureoverload playtest, loading  ~900 sprites of various sizes:

* chrome: 1113 --> 74 ms
* safari: 3697 --> 1273 ms

@Stanko can probably make a better torture test though, would be interested in test using spacedeck x's spritesheets with repacking enabled

test is failing because pupetteer is broken, idk why

* [X] Changeloged


Stanko:
Amazing work! 

I don't know why the old algorithm struggled on my sprites. Safari took 50 seconds to pack them. As of now I have 99 sprites and here are the results:

| Browser  | Alpha 27.1      | New       | My Vite plugin (both branches) | 
| -------- | --------------- | --------- | ----------- |
| Chrome 6 | 5 000 - 7 500     | 200 - 250 | 150 - 200   | 
| Firefox  | 23 000 - 25 000 | 200 - 250 | 100 - 150   | 
| Safari   | 45 000 - 50 000 | 500 - 600 | 300 - 350   |

This is an insane upgrade and with this I can even remove the Vite plugin (simpler the better, right). Plugin is still a bit better, and I'll keep it until this is released and tested a bit more.

Please note that I'm testing on MacBook with M3 Pro chip and 36 gigs of RAM.

dragoncoder047:
> Please note that I'm testing on MacBook with M3 Pro chip and 36 gigs of RAM.

I am testing on an M3/16G RAM and also have no clue why Safari is so much slower. You would think the first-party OpenGL->Metal translation would be the fastest, but apparently not.

dragoncoder047:
> I don't know why the old algorithm struggled on my sprites. Safari took 50 seconds to pack them. As of now I have 99 sprites and here are the results:

fwiw @Stanko this was because I used a brute force implementation that checked every other sprite for collisions which is $O(n^2)$ but what I did now uses a hashgrid to check collisions so it's approximately $O(n\ \log\ n)$ now.

ChatGPT is saying it's $O(n^3)$ to $O(n^2)$, but that's still an improvement.

lajbel:
Any way to split up? File changes are too large to review @dragoncoder047

amyspark-ng:
is this ready for review... @dragoncoder047

amyspark-ng:
I think file changes are not too large to review now so im gonna very rudely assume this is ready for review now, if i'm in the wrong mark as draft again

PR #1106, opened 2026-06-09
fix: preserve viewport aspect ratio in fullscreen, fix touch pos, and inspect resolution

imaginarny:
It follows up on the changes from #1096 and its `canvasToViewport()` calculation.

Before that, touch coords were way off in fullscreen, when canvas is set to fixed size and letterbox disabled. Although, fullscreen is aspect-fit resized so letterboxing is still present there and `onResize()` and `updateViewport()` don't account for that.

Mouse pos was correct because it scaled the viewport back to the correct ratio **on each mouse move**. After the mentioned PR, touch coords were only off because it calculated with as if stretched viewport instead (since mentioned functions skipped it). 

The mouse correction code was removed and viewport code was updated instead to scale and recalculate the ratio in that case as well which happens **only when resize happens**. So both mouse and touch are calculated with the viewport already updated with the correct ratio and scale. We now also differentiate between the screen and buffer viewports, more info in [the comment](https://github.com/kaplayjs/kaplay/pull/1106#issuecomment-5286778498).

🔥 Unexpected bonus: fixed the inspect resolution along at last, since the correctly updated viewport now allows that

## Summary

- Debug inspect is now really drawn unscaled as it was always meant to
- Fixed touch pos for the remaining `kaplay()` opt setup where `width` & `height` are defined, and `letterbox` disabled. Other options are basically unaffected/behave the same.
- Added `lockResolution` opt to be able to lock the canvas buffer resolution to the defined size and scale when resized by `letterbox`

### Testing

There are many playtests for those, which start with `mouseXY`. I updated those so they debug some setup info like resolutions, and added three more:
- `mouseSizeFixed` - with kaplay opts mentioned
- `mouseSizeAligned` - same, but when canvas is not in the original pos
- `mouseLetterboxLockRes` - `lockResolution` opt in action (canvas buffer size stays the same despite resized)

Will mention how to test touch here as well, that is also included in those two files:

> To test touch, enable touch simulation in dev tools responsive design mode, then press and hold mouse down and move around. Ensure that the red dot follows correctly in fullscreen (F).

- [x] Changeloged


imaginarny:
~~Sidenote, I trigger `resize` event on init if canvas dimensions don't match the options set as it usually happens with letterbox (but not only) and `resizeObserver` wouldn't catch these cases since~~ no need for that anymore, gl now has correct initial dimensions, same as the initial size is already canvas `offsetWidth/Height` (has to). I removed `updateCanvasScale()` call in init and `resizeObserver`, since it would be called thrice at times, where the ultimate one was always in the `onResize()` that now works as should. @Stanko, can you confirm your use case still works?

Stanko:
Hey, from looking at the changes I think they do, but I won't be able to test in next day or a two. As soon as I can, I'll send my findings here.

Stanko:
Actually there are two things that I found. The first one is a bug, because removing it on init assumes the initial scale is always 1, which is not the case if the canvas is already resized using CSS (like in my game). 

The second thing is that the resize observer schedules the resize on the next input tick (similar to our discussion in the PR for releasing input on blur) which leaves stale state. After a canvas size change (triggered by changing CSS without resizing the window), the first mouse move can run `canvasToViewport()` before resize handler updates the canvas scale.

Please bear in mind that I probably don't have the same understanding of the engine as you do, but from my experience it is good to be very careful about clearing up stale state. If you don't think this will have a huge performance impact I would keep it.

imaginarny:
Getting back to this as I was trying to come up with an example but I need more details on your use case @Stanko. Because this is more like it made canvasScale values redundant since it was as if fixing the symptoms unlike the core underlying issue. Will try to explain what lead to it:
- Before, besides the letterbox, it was possible to set the game to stretch (this was removed long ago since why would anyone do that, caused more issues and confusion than value added)
- Letterbox was then handled more strictly when either opt was enabled, letterbox or stretch
- After stretch gone, letterbox was kept handled strictly but some logic should have been updated
- Fullscreen was never working properly with letterbox, not resizing viewport in to aspect-fit the screen but keeping letterbox visible when it could be resized more for it to disappear, and canvas virtual size staying smaller than screen as well (not resized at all), more like zooming in instead
- Mouse input pos was patched converting on each move instead from screen to wrong viewport, and without letterbox specified, still as if it was stretched instead of aspect-fit, hence the letterbox effects needed still being applied only if opt enabled specifically
- That all resulted in many issues, notably out of sync viewport, incorrect coords calculated, and most importantly canvas not being resized properly thus your logical need for canvasScale

Now that the viewport and canvas sizes are updated correctly (and both are resized with aspect-fit), it means canvasScale is always going to be 1. Even if you resize dom canvas with css. So I’m not sure if there is something about your setup that's different or if there was a different reason/need or expected result I don't see. The only difference is in the game framebuffer resolution vs the viewport/canvas size, and logical canvas size is different than dom size only if pixelDensity set (but this won't affect canvasScale anyway). And your issue was that canvas logical size wasn't updated when canvas got resized (or went fullscreen), if I don't get it wrong. I hope it makes sense cause it's so mind-bending that even I'm getting lost (re)writing it (for the third time)..

Stanko:
This was my issue, click coordinates were calculated against the natural canvas size:

Screenshot 2026-07-30 at 09 43 29


So I think what you are saying makes total sense, and to test it you can add to your test:

```
k.canvas.width = 900;
k.canvas.height = 900;
```

as you can see, [on playground using alpha 27.1](https://play.kaplayjs.com/?code=eJxNjsFqAjEQhu95iumeZmFZ9tJCK3pSLyKIlx5KDyGZusuGjGSiRcR3bzZRbE7zf3z%2FTJRhLxFGmMOoj05f8Aq%2Fg439B7x1XQM9DYc%2B5gC3eqaUGluj%2FVlLm7XUe%2B%2B62ZOWwgOrsj6QXfIEtbX4pSA97U3PAStDPlKo6iZTMwTjCF8fkV1y9qvlPR9Z8D7qQHqav6dPsd%2FySWjLZ8LkwHwB12yVw21mEPmTg7OTkUq3UtzQZRdIBKufqgGsn12huD45JyYQeXwZ5F%2Bqy4I%2FZ1FcOQ%3D%3D&version=4000.0.0-alpha.27.1) it is showing red dot at the wrong position:

imaginarny:
That's not the case though. With this example besides the dom size you changed only the canvas resolution while you still enforced smaller picture size with css so you are downscaling that 900 (that draws 600 viewport) to 600 (so even smaller). You would also have to do:
```js
k.canvas.style.width = "900px";
k.canvas.style.height = "900px";
```
You can see that without it it's broken on master too where your canvasScale fix is included. But it doesn't matter, it isn't valid having those out of sync like that (rendering higher than css size) and I don't think you have it like that in your case either. I think you only change css size which is fine, although without this PR it would become blurry (when crisp not set).

Without the fix, the gfx viewport without the letterbox opt still stays at 600px no matter what, which obviously shouldn't once you do change the css size after or enter fullscreen. All should be in sync, else issues arise. This was your issue and why you did canvasScale, but it's not like that anymore.

For html canvas, the width and height attributes also work as its internal resolution. So its size should be synced with css size (screen size when in fullscreen), which also affects the resulting gfx viewport size (that will be aspect-fit and can be sub-rect of canvas/css size still) while the game virtual framebuffer is still at resolution set by opt. Otherwise it creates all those issues that this PR also fixes. Problem is that so far the other fixes and conversions were done on the faulty setup in some cases (mostly when letterbox opt isn't used or in fullscreen). As I summed up what lead to it before.

So I still need to find an example where canvasScale could turn different than 1 or case when it logically being now always 1 to be an issue to decide on how to move on with it. Would help if you could try this branch with your game if possible `npm i github:kaplayjs/kaplay#fix-fullscreen-touch-pos`. 🙏

Stanko:
Ah sorry, I was eager to test it. Long story short I just tested my game with this branch and it works as I would expect :)

imaginarny:
Oh my god, I haven't heard news that good that made me so relieved as this one in a long time. 😅 Although, now it means I can phase out the canvasScale altogether, but hopefully, I won't break anything else doing so.. 😄

Stanko:
Actually 🙈 there is an issue. If I turn on the debug mode, the inspect doesn't work until I resize a window. Only after resize it starts working.

Here is the video, you'll see that inspect starts working properly only after resize:

https://github.com/user-attachments/assets/e6f2feb9-a97c-4c71-962a-a15420ce47bd

Before the resize, you'll see this frame too, when it shows an overlay, showing "fighter" object which I'm not hovering at that point (fighter is the green/purple kamikaze ship)

Screenshot 2026-07-30 at 14 19 04

imaginarny:
Should be fixed now, can you try? 🤞 The setup like that was before just seemingly supported (hence the issues you ran into and other revealed xD), but now it should init with correct size.

Stanko:
I found a big issue that I missed earlier because I didn't realize that was changed too. 

I don't want my canvas size to be resized by the engine. My game's native resolution 480x270 and I render it at 960x540. This branch resizes the canvas breaking the pixel art aesthetics. I think it would be good to have a flag for a fixed sized canvas when creating a game.

|this branch|alpha 27.1|
|-|-|
|Screenshot 2026-07-30 at 20 12 20|Screenshot 2026-07-30 at 20 12 03| 

On a positive side, mouse events are fixed on init :)

imaginarny:
Canvas changing attr size should be the same as changing the viewport size, the game is still rendered at resolution you set, its then just drawn to canvas. Either downsized when buffer texture drawn or by browser. When downscaling pixelart, you are going to see squashed rows depending on where on the screen it is positioned and if there are odd numbers, can't avoid. Like if you just moved the whole canvas 1px down without resizing, the results would swap or you will see different results moving it pixel by pixel but you would see those results on both of them, just at different time/positions. Or when looked in different browser of different height of title/urlbar and etc. This is how I see them for example at 422px:

| this branch | alpha 27.1 |
|--------|--------|
| obrázok | a27 | 

On alpha 27.1 I actually see the Card Library top row cut off as well, but screenshot tool captured it better.

I don't know what can be done about it. Gotta think and test. It would undo almost the whole branch. Conversion issues, fullscreen is a whole different story, can't have properly scaled inspect in that case either, and so on.. 😔

Stanko:
It sure is complicated. Issue in my case isn't when the canvas gets too small, I just resized so the inspector is visible. 

When I need for canvas to stay fixed size is on viewports larger than the natural size of the game, on hi-dpi screens. In these cases there is enough physical pixels to draw everything crisp. From my limited testing (5-6 screens, Linux and MacOS) just letting the browser (or os?) works quite well)

Let me try to make an example:
- Natural width - 480x270
- Canvas size - 960x540
- Viewport - 640
- DPI - 3 

That gives us enough physical pixels (1920px) to render every game pixel as 4x4 (and canvas as 2x2) keeping the pixel art untouched.

If we resize the canvas to viewport, we are working with 640px canvas giving us 1.33x1.33 pixels per game pixels making the art blurry although there is enough pixels on the screen.

It is an artificial example, but I hope it makes sense.

EDIT: Every game engine I tried makes jump through a lot of hoops to render pixel art properly. Funny that low resolution games are complicated today 🙈 

EDIT2: Sorry for making your life harder 😅

amyspark-ng:
marking as still being actively worked on, correct me if i'm wrong

imaginarny:
It is since its draft, no need to request review yet either.

amyspark-ng:
makes sense yeah

imaginarny:
Pixelart being problematic to set up correctly is that nothing here done or said is correct in the end xD.. Rendering would have to be locked to whole pixels at given resolution and to get the uniform pixel sizes you would have to resize by whole pixels as well, no 100% of screen, so jumping x1, x2.. Otherwise it gotta interpolate with pixel density that is what you are missing, and those float dimensions are going to get crushed. Scale is just turning "source image" bigger. You just found an acceptable size/setup for your screens with it since you still do 100% as well, which really wasn't meant with fixed size kaplay setup so it's so cumbersome to deal with.

Also, you dont use device-width meta viewport tag so your width stays at 980 min. Altho that probably has no visible effect on the result, just mentioning how many factors are on top of it being in browser as well besides "modern" rendering vs pixelart xD.

So, I've been testing a ton, same with revert PR as well, but I gotta ask you @Stanko for one more test 🙏, with this branch still as is, but your setup without `scale` set but `pixelDensity` instead.. as simple as `pixelDensity: 2` could do and be basically what you are aiming for. Even if I usually recommend `Math.min(2, devicePixelRatio)` universally as only hi-dpi screen needs it and 2 is usually enough (if original res is already high). You might need higher, or play with the calculation after.

dragoncoder047:
imo the debug inspect and debug outline should not be affected by global scale or local transform, there should be options for debug scaling separately

Stanko:
Wait, it seems I messed up and tested the wrong branch. Give me 5 minutes 🙃 

Forget my previous comment, that was `master` branch instead, sorry once again. 

On this branch, even with pixel density it is not what I was aiming for. Long story short, I don't want canvas to be resized by the engine without me being able to control it. 

And please please please don't take this the wrong way, but if this gets merged I'll either lock the previous commit or my own fork with this feature disabled. That doesn't mean you shouldn't merge it if you think this is the best for a general use-case.

Stanko:
Here is a quick comparison (the width of the rendered canvas is identical, but screenshot frame is different):

|master|this branch|
|-|-|
|Screenshot 2026-08-06 at 17 45 03|Screenshot 2026-08-06 at 17 43 53|

Even with a larger canvas, pixel art is getting more deformed, it is quite visible on text. I think the OS does much better job rendering when canvas resolution matches the art scaling.

imaginarny:
Was this at pixelDensity 2 or is your screen devicePixelRatio higher? Since browser is spreading subpixels better and those cant be avoided when 100% sized, to let browser interpolate for the most part we should then set the less damaging resampling/interpolation that texture is drawn with then and set it to `"linear"`. That was unexpected, right? xD

Also no worries, Im just trying to get all cases working as close as possible and have this canvas matching window viewport since it's just worth it. It's just hard to test, cause edits and cases are misleading at times and can't spam you for test each change. But ideal would be just so all can have nice pixel art setups without having to do the hoops as you said, also doing contain, client/canvas conversions broadly, and explaining guide in the end. So I appreciate your help. I don't want to bother you this much, but can you test it now again with the `pixelDensity` instead of `scale` and now the `filter: "linear"`? Would like to see the result/comparison of that before I start kabooming it. Just make sure it's really on this branch and those options. 🙏

Stanko:
Let me fix a bug in my game first, and I'll do a comparison matrix:
master / branch
scale 2 / dpr 2

I'm on macbook, and with resolution dpr is exactly 2

Stanko:
## Baseline

* width/height in config: 480x270
* canvas rendered at 1200x675
* real device dpr: 2
* master branch uses this commit https://github.com/kaplayjs/kaplay/commit/8bd9f147d00aa4c36d9609586a560b7a61060cb9

## Canvas size

||scale: 2|pixelDensity:2|
|-|-|-|
|Branch|1200x675|2400x1350|
|Master|960x480|960x480|

## Screenshots, filter: nearest

Again, on this branch pixel art looks choppy. `pixelRatio: 2` looks better, but it is a hefty 2400px canvas 😅 No visible difference on master between scale and dpr.

||scale: 2|pixelDensity:2|
|-|-|-|
|Branch|branch-scale-2|branch-dpr-2|
|Master|master-scale-2|master-dpr-2|

## Screenshots, filter: linear

Haha this is so blurry I can't tell the difference 🙈 

||scale: 2|pixelDensity:2|
|-|-|-|
|Branch|branch-linear-scale-2|branch-linear-dpr-2|
|Master|master-linear-scale-2|master-linear-dpr-2|

imaginarny:
Uffffffffff, I really did test with different changes and did typo with the filter cause this `[opt.filter ?? ctx.opts.texFilter ?? "nearest"]` bit in gl setup so it wasn't even used and I jumped conclusion too early and so so stupidly 💀. Since idea to let browser resample can't happen as it's still resampled by fb cause source resolution is still small... Huge fail. I give up. And I reverted those changes already that I tested. There on iPad it looked better than ever for me, eg I can tell on your version the arrow is on subpixels. Ok, don't mind me, my brain is boiling..

| master | branch |
|--------|--------|
| master | branch |

imaginarny:
We just learned that letting gpu resample with nearest really sucks since with linear blur was identical, haha.

imaginarny:
Ok, so I restored behavior when canvas is set to fixed size (no letterbox) and resized with CSS, the canvas buffer won't resize. It wasn't designed like that before on purpose since it was not expected to resize (altho fullscreen did and we are here), but it's a good option as we saw with the browser based "resampling" results instead of gl in some cases. So I also added a new opt `lockResolution` so users can still use `letterbox` (that makes canvas take 100% available space as well), but canvas buffer will stay at the defined size and scale. It's basically like not enabling letterbox but resizing canvas manually to 100% with CSS and `object-fit: contain`, which you do. Both will work. Contain is now set by default since size was always meant to dictate the ratio as well and we removed stretch way ago as I mentioned before anyway. `lockResolution` can be seen in a new `mouseLetterboxLockRes` playtest.

To achieve the correct viewport setup I now differentiate between the buffer viewport (original) and the screen viewport. That was essential and the culprit of it all, where the screen is the one that needs the calculation and conversions the most, and in our usual setup either both screen and buffer are equal or buffer one stays basically at the initial setup. Now, the "`lockResolution`" behavior made it more obvious, since it's local gfx space vs screen css space.

Just a note about the debug scaling, in this case it has to stay like that since when locked to the low res, there are not enough physical pixels to downscale it, it would become unreadable completely, obviously. In that case, only a separate canvas altogether would help.. Or when used with `letterbox`, `lockResolution` can be disabled for a brief testing purposes if needed.

Thank you for your cooperation and patience @Stanko 🙏. You can test now, hopefully it will be all good (should be). 🤞

Stanko:
This is perfect and works exactly how I would expect it to - game/art wise and the debug mode. Thanks for the effort! 🙌

lajbel (approved):

lajbel:
Wonderful ⭐

PR #1109, opened 2026-06-15
fix: make pos, skew and scale use a proxy

### LINKED ISSUE #1108: bug: modifying `PosComp.pos.x` and `PosComp.pos.y` doesn't move the object
### Bug Description

changing the position of an object after the its first update loop via `pos.x` and `pos.y` doesn't visually move the object.

### Version

present in v4000.0.0-alpha.26 to (as of writting) v4000.0.0-alpha.27.1

### Playground Link

[playground](https://play.kaplayjs.com/?code=eJxtjrEOwjAMRHd%2FxY22qNryAeEvmBBDmmShVRI1QSiq%2BHdMWLnhdNL5WUerzZttLEQuxVKRlkdwFQbWe74RVDkVlqHHXTs%2Bz%2FMANaG7Yj9g1KOx4WS%2BBabplfa1UIrX7G0NzAJzwdF%2F%2FAfQCSxPnRC3puYCveUD6h0tSA%3D%3D&version=4000.0.0-alpha.27.1)

``` typescript
kaplay()

const object = add([
    pos(),
    rect(100, 100)
])

object.pos.y += 100 //works
onUpdate(() => {
    object.pos.y += 100 // works but only once
})
```

### Extra information

it you prints out the object position, it actually changes so this is only a visual bug.

methods like `moveTo` and `moveBy` still works and setting `pos` directly also works.

### Summary

- [ ] Fixed in v4000
- [ ] Fixed in v3001

ErikGXDev:
To make code like `obj.pos.x += 10` work, I have rewritten the pos component to use a proxy, which enables this. This was not possible before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,7 +61,7 @@ So your change should look like:
   time (#1082) - @nojaf
 - Fixed mouse coordinates not being calculated properly when canvas is resized
   by CSS and wasn't rendered at its natural size (#1096) - @Stanko
-- Modified `pos`, `skew` and `scale` components to make operations like `+=` or `-=` work again - @ErikGXDev
+- Modified `pos`, `skew` and `scale` components to make operations like `obj.pos.x += 1` work again (#1109) - @ErikGXDev
 
 ## [4000.0.0-alpha.27.1] - 2026-05-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ So your change should look like:
   time (#1082) - @nojaf
 - Fixed mouse coordinates not being calculated properly when canvas is resized
   by CSS and wasn't rendered at its natural size (#1096) - @Stanko
+- Modified `pos`, `skew` and `scale` components to make operations like `+=` or `-=` work again - @ErikGXDev
 
 ## [4000.0.0-alpha.27.1] - 2026-05-12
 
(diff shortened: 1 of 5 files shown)


ErikGXDev:
made a small update to use Reflect which seems to be a better way of setting the position

ErikGXDev:
I have also added proxies to scale and skew (and have updated the title accordingly), so those things will also work like pos now (and I also messed up the commit message but I fixed it)

imaginarny:
Can you add [bouncingBeans.js](https://github.com/user-attachments/files/29009905/bouncingBeans.js) to `tests/playtests` folder? It's based on the demo you shared. Also, run `fmt` again and update the changelog, pls 🙏 . I've been testing it in multiple browsers/devices and there is no change in performance. 🙂

ErikGXDev:
I added the example and formatted the files 👍. I added it to the "fixed" category in the changelog as well

CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ So your change should look like:
   time (#1082) - @nojaf
 - Fixed mouse coordinates not being calculated properly when canvas is resized
   by CSS and wasn't rendered at its natural size (#1096) - @Stanko
+- Modified `pos`, `skew` and `scale` components to make operations like `+=` or `-=` work again - @ErikGXDev
```
imaginarny:
```suggestion
- Modified `pos`, `skew` and `scale` components to make operations like `obj.pos.x += 1` work again (#1109) - @ErikGXDev
```

Would change it slightly so there's a clear example (same as before) that it's meant for its props. Also included #PR. Sorry for annoyances.

imaginarny (changes requested):

ErikGXDev:
Okay, I added the change

imaginarny (approved):
Thank you Erik 🙂. Can't think of anything where it would have a negative effect.

PR #1112, opened 2026-06-25
feat: add nextFrame helper function

imaginarny:
The `nextFrame(() => {})` is basically the `wait(0, () => {})`. Recently, a few more use cases have emerged when deferring to the next frame would make sense or is required. This alternative function makes its use more obvious rather than seemingly hacky `wait(0)`. Some of those are documented in types/JSDoc.

Playtest prevents the one where if you register an event in the same event listener, it would get triggered immediately, since it's added to the listeners stack before the current frame/loop finishes processing those listeners.

## Summary

Adds `nextFrame()` helper function to defer/run a function on the next frame.

- [x] Changeloged

### DIFF
--- a/src/ecs/components/misc/timer.ts
+++ b/src/ecs/components/misc/timer.ts
@@ -152,6 +152,7 @@ export function timer(maxLoopsPerFrame: number = 1000): TimerComp {
             this: GameObj<TimerComp>,
             action?: () => void,
         ): TimerController {
+            // will run only once
             return this.loop(0, action ?? (() => {}), 1, true);
         },
         tween<V extends LerpValue>(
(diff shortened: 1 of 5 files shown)


src/ecs/components/misc/timer.ts
```diff
@@ -144,6 +148,12 @@ export function timer(maxLoopsPerFrame: number = 1000): TimerComp {
         ): TimerController {
             return this.loop(time, action ?? (() => {}), 1, true);
         },
+        nextFrame(
+            this: GameObj<TimerComp>,
+            action?: () => void,
+        ): TimerController {
+            return this.loop(0, action ?? (() => {}), 1, true);
```
lajbel, line 156: I would comment here it will just run once. First seeing looks weird

lajbel (approved):

src/ecs/components/misc/timer.ts
imaginarny: Added `// will run only once` comment. Can merge?

lajbel (approved):
LGTM, ok to merge

PR #1115, opened 2026-06-29
fix: align styled-text style positions with grapheme clusters

mahirhir:
`compileStyledText` builds `charStyleMap` keyed by `renderText.length`, i.e. UTF-16 code units, while emitting the text one code unit at a time. `formatText` then splits the rendered text with `runes()` and applies styles with `charStyleMap[cursor]`, where `cursor` is a grapheme index.

For plain ASCII those two indexings match, but they drift apart as soon as the text contains a character that is more than one code unit: an emoji, a ZWJ sequence, a skin-tone or flag sequence, or an astral-plane CJK ideograph (CJK Extension B, e.g. names written with 𠮷). After such a character every following style is shifted, so it lands on the wrong grapheme or is dropped.

A couple of concrete cases:

- `😀[c]x[/c]` keys the style at `2`, but `runes("😀x")` puts `x` at grapheme index `1`, so the style is lost.
- `[c]😀b[/c]` produces keys `{0,1,2}` for a string `runes()` sees as two graphemes.

The fix makes `compileStyledText` walk grapheme clusters with the same `runes()` helper `formatText` already uses, keying `charStyleMap` by grapheme index. The input is normalized to NFC up front (which `runes()` does internally too) so the slice lengths stay consistent. Escapes (`\x`) now also consume a whole grapheme rather than a single code unit.

Added `tests/auto/styledText.spec.ts` covering ASCII (unchanged), an emoji, an astral CJK character, and a ZWJ family sequence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,10 @@ So your change should look like:
   e.g. focused `textInput` now use `preventDefault()` (#1114) - @imaginarny
 - Updated `drawSprite()` to allow slice9'ed sprites, and so now the `sprite()`
   component just calls `drawSprite()` (#1036) - @dragoncoder047
+- `StyledTextInfo` now carries the rendered text's grapheme clusters as a
+  `runes` array, and `compileStyledText()` takes an optional `locale`
+  parameter, so `formatText()` no longer segments the same string twice
+  (#1115) - @mahirhir
 
 ### Fixed
 
@@ -98,6 +102,8 @@ So your change should look like:
   duration (#1117) - @imaginarny
 - Fixed triangulate by updating the convexity of nearby vertices after removing
   a concave vertex during ear cutting (#1134) - @mflerackers
+- Fixed styled text (`[tag]...[/tag]`) style positions desyncing after an
+  emoji, ZWJ sequence or astral character (#1115) - @mahirhir
 
 ## [4000.0.0-alpha.27.1] - 2026-05-12
 
--- a/src/core/contextType.ts
+++ b/src/core/contextType.ts
@@ -6696,14 +6696,15 @@ export interface KAPLAYCtx {
     formatText(options: DrawTextOpt): FormattedText;
     /**
      * Parses the text that has formatting tags, and returns the unstyled text
-     * (the actual characters that will be displayed) as well as which styles are
-     * active on each character.
+     * (the actual characters that will be displayed, both as a string and as
+     * an array of grapheme clusters) as well as which styles are active on
+     * each character.
      *
      * @since v4000
      * @group Rendering
      * @subgroup Text
      */
-    compileStyledText(text: any): StyledTextInfo;
+    compileStyledText(text: any, locale?: string): StyledTextInfo;
     /**
      * Create a canvas to draw stuff offscreen.
      *
--- a/src/gfx/formatText.ts
+++ b/src/gfx/formatText.ts
@@ -6,7 +6,7 @@ import { Color, rgb } from "../math/color";
 import { vec2 } from "../math/math";
 import { _k } from "../shared";
 import type { Outline } from "../types";
-import { firstRune, runes } from "../utils/runes";
+import { firstRune } from "../utils/runes";
 import { alignPt } from "./anchor";
 import type { FormattedChar, FormattedText } from "./draw/drawFormattedText";
 import type { CharTransform, DrawTextOpt } from "./draw/drawText";
@@ -31,6 +31,13 @@ export type FontAtlas = {
 export type StyledTextInfo = {
     charStyleMap: Record<number, [string, string][]>;
     text: string;
+    /**
+     * The rendered text split into grapheme clusters, in render order.
+     * charStyleMap keys index into this array. Carrying it here lets
+     * formatText() reuse the segmentation compileStyledText() already did
+     * instead of running runes() over the same string a second time.
+     */
+    runes: string[];
 };
 
 function applyCharTransform(fchar: FormattedChar, tr: CharTransform) {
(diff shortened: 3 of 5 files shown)


dragoncoder047 (changes requested):
If you're going to use runes() like this, you should change it to have a parameter to just get the first rune and not all of them, otherwise the runtime will be $O(n^2)$ (up from $O(n)$!!) since it is now trying to split the entire remaining string into runes for every character in the string.

Also, using `const { 0: grapheme } = ...` instead of `const [grapheme] = ...` is faster and uses less memory since it doesn't create an iterator.

mahirhir:
Good catch. I added a firstRune() helper next to runes() that pulls just the first grapheme from the Intl.Segmenter iterator and returns on the first segment, so it no longer splits the whole remaining string on every character and the walk is back to O(n). compileStyledText now calls firstRune(text) directly, so there is no array or destructuring at the call site either. It falls back to runesLegacy()[0] where Intl.Segmenter is unavailable.

dragoncoder047:
I just thought of another optimization: what is effectively being done is that compileStyledText is chopping runes off incrementally, and then formatText is calling runes() again, so it's duplicating the work. Maybe the `StyledTextInfo` can be made to have an array of runes instead of a string which has to be cut up again.

mahirhir:
Yeah, that duplication is real. `compileStyledText` segments the string to pull each grapheme off, and then `formatText` runs `runes()` over the same string again, so we're segmenting twice over the same text.

Carrying a rune array on `StyledTextInfo` would remove that outright. My only hesitation is that it changes the shape of a type other call sites read, so it's a wider change than the alignment bug this PR is fixing, and it'd probably be easier to review on its own.

I'm happy either way: fold it in here, or land this fix and do the rune-array change as a follow-up. Which would you prefer?

dragoncoder047:
me personally would be happy either way but rolling it in one would probably be best to reduce noise, besides, v4000 still isn't officially released yet so large API changes are still fair game (I did that with the texture frame format already and it broke a lot of stuff, but not in a way that couldn't be fixed easily)

mahirhir:
Rolled it in. That is commit `56433fc`, pushed the day after your comment, and I never said so here, so this PR has been sitting waiting on a message rather than on code.

What changed:

- `StyledTextInfo` gains `runes: string[]`, the rendered text in render order. `charStyleMap` keys index into it, which is what they already meant.
- `compileStyledText(txt, locale?)` takes the locale, so its segmentation matches what `formatText` used to ask `runes()` for.
- `formatText` destructures `runes: chars` out of the compile result instead of calling `runes(text, opt.locale)` on the same string. The text is segmented once now, and the emit loop still uses `firstRune`, so the walk stays O(n).

The other consumers of the type only read `charStyleMap` (`drawFormattedText.ts` takes `StyledTextInfo["charStyleMap"][number]`, `context.ts` just re-exports the function), so the added field is additive for them.

Ran here: `tests/auto/styledText.spec.ts` passes 7 of 7, and `tsc --noEmit` is clean. I had to run vitest with `--environment node` for that file, because the repo default is the puppeteer environment and I have no browser on this machine, so I have not exercised the rest of `tests/auto`.

CHANGELOG carries both entries, the fix and the `StyledTextInfo` change, since the second one is a shape change to a public type.

mahirhir:
Marked ready for review. The follow-up you asked for went in as `56433fc` on 31 August and I never said so here, which is why this sat as a draft for a week with the work already done.

On rolling it into one: I have kept it as the single PR you preferred rather than splitting it.

What I verified before flipping it, so you do not have to take my word:

- `styledText.spec.ts` 7/7. The repo default runner is puppeteer, so I ran it with `--environment node`; if that matters to you I can add the browser run instead.
- `tsc --noEmit` clean.
- Other readers of `charStyleMap` only read it, so the position change does not reach them.

The duplicate-work point you raised in July, that `compileStyledText` chops runes incrementally and then `formatText` calls `runes()` again, is real and I have not addressed it. It is a separate change and I did not want to fold a perf refactor into a correctness fix. Say the word if you would rather have both at once.

amyspark-ng:
Since to be ready for review, thank you for your service

mahirhir:
Thanks. Since this has been sitting, here is the state of `master` today rather than a restatement of the PR, and one thing I had not spelled out before.

The branch is only 7 commits behind `master` and still `MERGEABLE`, so drift is not the blocker.

The alignment problem is visible in two line numbers on `master`. The style map is written keyed by a UTF-16 offset:

```ts
// src/gfx/formatText.ts:67, inside compileStyledText
charStyleMap[renderText.length] = styleStack.slice();
```

and read keyed by a grapheme index:

```ts
// :312  chars is a grapheme array
const chars = runes(text, opt.locale);
// :345
while (cursor < chars.length) {
// :385
styles: charStyleMap[cursor] ?? [],
```

`renderText.length` counts UTF-16 code units. `cursor` counts graphemes. Those two agree only while every grapheme happens to be one code unit:

```
text        utf16  graphemes   last-grapheme key
"abc"           3          3   agree
"aあb"          3          3   agree
"a😀b"          4          3   DIVERGE 3 vs 2
"aéb"          4          3   DIVERGE 3 vs 2   (e + combining acute)
"a👨‍👩‍👧b"    10          3   DIVERGE 9 vs 2
```

(That table is Node with `Intl.Segmenter`, measuring the premise only. It is not a run of kaplay.)

So a style applied after an emoji or a combining mark lands on the wrong character, and the further into the string the styled run is, the further it drifts. It is invisible with ASCII, which is why it survived.

There is a second, quieter divergence in the same two lines. `compileStyledText(txt)` takes no locale, while `runes(text, opt.locale)` does. The indices are produced by a locale-less segmentation and consumed by a locale-aware one, so for a locale where the two disagree about boundaries the keys are wrong even without any surrogate pairs.

The follow-up @dragoncoder047 asked for in July is what fixes both at once, and it is already in this branch as `56433fc`: `StyledTextInfo` carries `runes: string[]`, `compileStyledText` takes the locale, and `formatText` destructures that array instead of calling `runes()` on the same string again. One segmentation, one index space, keys that mean what they are read as.

What I have and have not got:

```
styledText.spec.ts   7/7   (local, --environment node; the repo default runner is puppeteer)
tsc --noEmit         clean (local)
CI on this PR        only the `greeting` job ran
```

There is no test or lint run on this branch to point at, so the local numbers are mine rather than yours. Happy to rebase onto current `master` if a fresh run would help, and happy to split the rune-array change back out if one PR is now doing too much.

PR #1117, opened 2026-06-30
fix: finish tween only once and clamp time to duration

imaginarny:
It fixes two issues based on a previous report on DC. If tween has finished and you call `finish()` at the same time or again after, it will set to `to` value and run all `onEnd` functions again. Even if it has already finished. Second issue comes from a workaround, having to check beforehand the `timeLeft` value, which instead of being `0` was a negative number. So finish code was unified and ensured it will only run once, with no need to check `timeLeft`, and the time values were clamped to the tween duration.

## Summary

Fixed `finish()` being able to complete an already finished tween, and prevented `timeLeft` from going negative and `currentTime` from exceeding the duration.

- [x] Changeloged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,9 @@ So your change should look like:
   `obj.pos.x += 1` work again (#1109) - @ErikGXDev
 - Fixed `isKeyDown` and `isButtonDown` getting stuck on game loosing focus
   (#1101) - @Stanko
+- Fixed `finish()` being able to complete an already finished tween, and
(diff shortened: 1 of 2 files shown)


amyspark-ng:
does finish() assign the property to the "to" param?

imaginarny:
Does on tween completed or when called `finish()` before it completes. Just can't finish an already finished tween again, it does nothing in that case.

lajbel (approved):

lajbel:
Great, I think we had this bug around for a long time

lajbel:
Please resolve conflicts @imaginarny

PR #1118, opened 2026-07-03
docs: add testing guidance

CEREBR4L:

Not tied to a specific issue. Testing isn't documented anywhere right now,
`CONTRIBUTING.md` doesn't mention it at all, so it's a guess for anyone
contributing whether a change needs a vitest spec, a playtest, or nothing.
I ran into that gap myself while trying to get more involved here, and
figured writing down what's actually going on (a few different tools for
different situations, most `tests/playtests` files aren't even tagged
`@test` on purpose, etc.) would give everyone, myself included, the same
shared understanding instead of each contributor re-deriving it on their own.

## Summary

- Added `TESTING.md`: the three test tiers, where each kind of fix goes, and
  when it's fine to skip a test.
- Linked it from `CONTRIBUTING.md`, and added `pnpm run test` / `test:vite`
  to the pre-push checklist.

One thing I want to float rather than just decide: the doc suggests saying
why you're skipping a test, even if it's just "visual change, verified with
`pnpm dev`". Open to cutting that if it's not something you want to
normalize. As does quietly add the stupulation all PRs should at least 
address testing.

Also aware #1116 is reworking testing too (Playwright, `tests/e2e`, a new PR
template with reviewer-side test checkboxes). This doc will need a follow-up
once that lands, or I'm happy folding it into that PR instead if you'd
rather not have two testing changes in flight at once.

- [ ] Changeloged

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,12 +45,19 @@ Normally the definitions of components, types, interfaces (i.e: `SpriteComp`,
 Your best option is use **ctrl + click** on the type to go to the definition and
 see the JSDoc comments. Other option is to use the search feature of your IDE.
 
+## Testing
+
+See [TESTING.md](TESTING.md) for what we test, where each kind of test goes, and
+when it's fine to skip one. Tests aren't required for every PR, if you skip one,
+a short reason in the PR is enough.
+
 ## Commiting your changes
 
 1. Follow our [conventional commits](#conventional-commits-guide) format. You
    can see how seeing the commit history.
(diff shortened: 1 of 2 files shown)


lajbel:
I will mark it as 'draft' since current work in test PR, will notify you when that is done

lajbel:
Hi @CEREBR4L, PR is ready #1156

amyspark-ng:
@CEREBR4L marked as waiting for author, we're currently using npm rather than pnpm, please update and mark your PR as ready for review so our team can review it

PR #1119, opened 2026-07-03
fix: resolve gamepad mappings by multi-tier matching

### LINKED ISSUE #1110: bug: custom extended gamepad buttons don't work anymore
### Bug Description

i added a mapping for the ps5 controller in #867 but it doesn't work anymore, since the id property has changed, so the touchpad is back to "undefined":

old: `"DualSense Wireless Controller (STANDARD GAMEPAD Vendor: 054c Product: 0ce6)"`

new: `"DualSense Wireless Controller (STANDARD GAMEPAD)"`

I can't test the other ones, but I bet they changed too.

So we need another way of detecting custom controller mappings.

### Version

master

### Playground Link

_No response_

### Extra information

_No response_

### Summary

- [ ] Fixed in v4000
- [ ] Fixed in v3001

CEREBR4L:

Fixes #1110.

`Gamepad.id` isn't standardized by the Gamepad API spec - its exact text is whatever the browser decides to report, and it's not guaranteed to stay the same between browser versions. Chrome's id for the DualSense used to include `Vendor: 054c Product: 0ce6`, and no longer does, which is why the mapping added in #867 stopped matching and the touchpad button went back to `undefined`. The underlying issue isn't specific to DualSense though - we were matching gamepads by comparing the *entire* id string, which breaks any time a browser changes wording/tagging/casing anywhere in that string, for any controller, not just this one.

## Summary

- [x] Changeloged

Re-keyed the built-in gamepad table to match on the vendor:product hex pair (parsed out of `id`) instead of the raw id string, since that's the one part of the id that's actually stable across the kind of text-format changes that caused this bug. Custom controller maps passed via `kaplay({ gamepads: {...} })` are untouched and still keyed by the literal id string, so this isn't a breaking change for anyone already using that option.

For the specific case in this issue - Chrome now omits vendor/product from the DualSense id entirely, so there's nothing left to parse - added a second, more conservative fallback: a substring match against the controller's own reported name, but only for controllers where that name text is actually distinctive enough to trust (`"DualSense"` is safe to match on; DualShock 4's reported name is just the generic `"Wireless Controller"`, which plenty of unrelated pads also report, so that one's deliberately not included). This is what actually recovers the touchpad button for the id in this issue.

While in there, also added DualShock 4 (`054c:05c4` / `054c:09cc`) as a built-in mapping - same touchpad-beyond-standard-range situation as DualSense, and it wasn't covered before at all.

Also surfaced a `controllerName` field on `KGamepad` (e.g. `"DualSense"`, `"Standard Gamepad"` if unrecognized) - display/debugging only, doesn't affect matching, but makes it possible to write tests (and games) against a readable label instead of the vendor:product hex.

### Testing

Added `tests/auto/gamepadId.spec.ts` - the id-parsing and matching logic is pure (no DOM/browser dependency), so it's covered directly with real `Gamepad.id` strings observed across Chrome/Firefox for DualSense, DualShock 4, Switch Pro Controller, Joy-Con, and a few controllers that are deliberately *not* in the built-in table (Xbox, a generic/knockoff pad) to confirm those still degrade to the default mapping instead of misidentifying as something they're not. Vendor:product pairs cross-checked against [SDL_GameControllerDB](https://github.com/mdqinc/SDL_GameControllerDB) to make sure none of it was guessed.

Also added tests/playtests/gamepadControllerName.js, a manual playtest for the parts of this that need a real gamepad in a real browser (can't be asserted in CI). Shows the resolved controllerName, and a debug panel that pairs the per-gamepad instance API (gp.isDown(), gp.getStick(), gp.getAnalog()) against the equivalent global functions (isGamepadButtonDown(), getGamepadStick(), getGamepadAnalogButton()) side by side so any mismatch is visually obvious, plus a buttons: {...} binding wired to a gamepad button to confirm the higher-level binding system resolves correctly too, not just the raw per-device events. Manually verified against a DualSense, a DualShock 4, and two Xbox-family pads (one official, one generic/knockoff). Lack hardware for further physical testing unfortunately. 

EDIT: For clarity claude llm was used to assist in this work, but the ideas, structure, and most of the code for implementation was mine. Testing used it more heavily but still reviewed. All changes have been reviewed and I can explain how this feature works in detail if needed. It is not "vibe coded" but LLM assisted.

### DIFF
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -3,6 +3,7 @@
 import type {
     Cursor,
     GameObj,
+    GamepadDef,
     KAPLAYOpt,
     Key,
     KGamepad,
@@ -34,6 +35,7 @@ import {
     releaseButton,
     setButton,
 } from "./buttons";
+import { resolveGamepadMap } from "./gamepadId";
 import {
     ButtonProcessor,
     type ButtonsDef,
@@ -59,7 +61,7 @@ export class ButtonState<T = string, A = never> {
     }
     process(state: AppState) {
         if (this._downEv !== null) {
-            this.down.forEach(b => {
+            this.down.forEach((b) => {
                 state.events.trigger(this._downEv as any, b, this._arg);
             });
         }
@@ -140,7 +142,7 @@ class FPSCounter {
         }
     }
     calculate() {
-        return this.fps = this.curNSamples / this.accumulator;
+        return (this.fps = this.curNSamples / this.accumulator);
     }
     ago(ago: number) {
         return this.history.at(this.i - ago - 1);
@@ -224,6 +226,9 @@ export const initAppState = (opt: {
         ),
         mergedGamepadState: new GamepadState(null),
         gamepadStates: new Map<number, GamepadState>(),
+        // resolved button map per gamepad index, computed once on connect
+        // rather than re-resolved from the raw id string every frame
+        gamepadMaps: new Map<number, GamepadDef>(),
         lastInputDevice: null as "mouse" | "keyboard" | "gamepad" | null,
         // unified input state
         gamepads: [] as KGamepad[],
@@ -263,8 +268,7 @@ export const initApp = (
 
     function updateCanvasScale() {
         const pd = opt.pixelDensity || 1;
-        state.canvasScaleX = state.canvas.width / pd
-            / state.canvas.offsetWidth;
+        state.canvasScaleX = state.canvas.width / pd / state.canvas.offsetWidth;
         state.canvasScaleY = state.canvas.height / pd
             / state.canvas.offsetHeight;
     }
@@ -307,7 +311,7 @@ export const initApp = (
 
     function screenshotToBlob(): Promise<Blob> {
         return new Promise<Blob>((resolve, reject) => {
-            state.canvas.toBlob(b => {
+            state.canvas.toBlob((b) => {
                 if (b !== null) resolve(b);
                 else reject(new Error("failed to make blob"));
             });
@@ -366,9 +370,11 @@ export const initApp = (
     }
 
     function isFullscreen(): boolean {
-        return document.fullscreenElement === state.canvas
+        return (
+            document.fullscreenElement === state.canvas
             // @ts-ignore
-            || document.webkitFullscreenElement === state.canvas;
+            || document.webkitFullscreenElement === state.canvas
+        );
     }
 
     const isFocused = () => {
@@ -470,7 +476,7 @@ export const initApp = (
     }
 
     function isTouchscreen() {
-        return ("ontouchstart" in window) || navigator.maxTouchPoints > 0;
+        return "ontouchstart" in window || navigator.maxTouchPoints > 0;
     }
 
     function mousePos(): Vec2 {
@@ -551,9 +557,7 @@ export const initApp = (
             );
     }
 
-    function getGamepadAnalogButton(
-        btn: KGamepadButton,
-    ): number {
+    function getGamepadAnalogButton(btn: KGamepadButton): number {
         return state.mergedGamepadState.analogState.get(btn) ?? 0;
     }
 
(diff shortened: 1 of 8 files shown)


src/app/app.ts
dragoncoder047, line 1: Most of the formatting changes here look like just formatting changes, can you revert them and/or run `pnpm fmt` to make sure it's formatted?

dragoncoder047 (changes requested):
The gamepad stuff seems like it could be simplified to use a regex for matching, so instead of `GAMEPAD_MAP` being a `Record<string, GamepadDef>`, it could be `Map<RegExp, GamepadDef>`. Though I do like the attempt at a cross platform parser for VID/PID.


As well as the `"name"` field (which is just the cosmetic human-readable name of the gamepad), you might take this opportunity to add a `"type"` field that has a machine-readable type like 

```ts
type GamepadType = "ps4" | "ps5" | "xbox" | "switch" | undefined;
```

Having KAPLAY provide this automagically would be really cool, and help you select the button image glyphs shown to the player.

[I was kind of doing something similar in Aelith on my own](https://github.com/r47onfire/aelith/blob/4d6b6d7c9fb2f04b017bc10903fa00a2fb5a5433/archive/src/controls/autodetectGamepad.ts#L1), but having KAPLAY do it for everyone would probably be pretty nice.

src/app/app.ts
CEREBR4L: This is the result I get from running `pnpm fmt` even when targeting the file directly with `pnpm fmt src/app/app.ts` directly. I have no config changes or anything locally just as I pulled it down so unsure why this is happening. Any suggestions?

src/app/app.ts
```diff
@@ -59,7 +61,7 @@ export class ButtonState<T = string, A = never> {
     }
     process(state: AppState) {
         if (this._downEv !== null) {
-            this.down.forEach(b => {
+            this.down.forEach((b) => {
```
dragoncoder047: You *must* have done something other than `pnpm fmt`, since dprint would never do this, I tested it.

CEREBR4L: This was zed's default on save formatting and I must have saved that file with zed as I worked. Wrongly assumed it was something like that file hadn't been touched since formatting rules were put it and never checked. Has since been fixed thank to some help in discord for others!

CEREBR4L:
> The gamepad stuff seems like it could be simplified to use a regex for matching, so instead of `GAMEPAD_MAP` being a `Record<string, GamepadDef>`, it could be `Map<RegExp, GamepadDef>`. Though I do like the attempt at a cross platform parser for VID/PID.

Had a think through this and I am not sure I agree fully, yes the engine code is simplified but in at the cost of making the API for users harder to grasp and I am not sure I can personally identify a use case for the above but happy to discuss. I have outlined my fully thoughts below:

The vendor:product pair is exact, enumerable data (real USB VID/PID values from SDL_GameControllerDB), not something that needs fuzzy/pattern matching. There’s no entry in the built-in table where we want “anything matching this shape,” only “this specific pair.”

Regex keys also means gamepad.json stops being plain JSON and turns into a .ts file as you pointed out on discord. Right now adding a controller is a one-line JSON edit; that’s the part of this file that gets touched most often, by contributors who have nothing to do with the matching logic itself, and I don’t want to make that “write/read a regex literal” instead of “paste a hex pair.” Same problem hits kaplay({ gamepads: {...} }) - regex object keys are a worse day-to-day experience than plain strings for what’s meant to be a quick per-game override.

The one place we do want pattern-style matching - browsers that strip vendor/product from id entirely - already has it, via the matchNames substring fallback. That’s deliberately narrower than full regex (see the DualShock 4 comment in gamepadId.ts: its reported name is generic enough that a careless pattern would misidentify unrelated pads), but it covers the same need at the right scope.

Open to it if there’s an actual matching case I’m missing, but I don’t think the built-in table needs to change shape to support this?

CEREBR4L:
> As well as the `"name"` field (which is just the cosmetic human-readable name of the gamepad), you might take this opportunity to add a `"type"` field that has a machine-readable type like
> 
> ```ts
> type GamepadType = "ps4" | "ps5" | "xbox" | "switch" | undefined;
> ```
> 
> Having KAPLAY provide this automagically would be really cool, and help you select the button image glyphs shown to the player.
> 
> [I was kind of doing something similar in Aelith on my own](https://github.com/r47onfire/aelith/blob/4d6b6d7c9fb2f04b017bc10903fa00a2fb5a5433/archive/src/controls/autodetectGamepad.ts#L1), but having KAPLAY do it for everyone would probably be pretty nice.

These are my thoughts on it, happy to revise if you see it differently.

`GamepadDef` (and `KGamepad`) now has a `type` field. It's derived from the same vendor:product hex `gamepadId.ts` already parses for the button table, not a separate detection pass: built-in entries in `gamepad.json` carry `type` directly, and for controllers that don't have (or need) a full entry - Xbox pads work fine under the "default" mapping, so there's no dedicated table row for them - there's a small vendor-id-only fallback (Sony/Microsoft/Nintendo) that kicks in instead of leaving those `undefined`.

One change from what you sketched: no PS4/PS5 split. DualShock 4 and DualSense report the same button iconography (face buttons, d-pad, sticks, Share/Options, touchpad), so a generation split wouldn't actually select between different glyph sets - just one `"playstation"` value.

The bit I want to flag specifically: `type` isn't a closed enum. It rides on `GamepadDef` same as `name`/`buttons`, so it's covered by `kaplay({ gamepads: {...} })` for free - and since we don't (and can't) ship a mapping for every controller that exists, `type` accepts any string, not just `"playstation" | "xbox" | "switch"`. So if someone's shipping Steam Controller support in their own game, they're not blocked waiting on us:

```ts
kaplay({
  gamepads: {
    "28de:1142": {
      name: "Steam Controller",
      type: "steam",
      buttons: {/* ... */},
      sticks: {/* ... */},
    },
  },
});
```

`gp.type` on their end just reads back `"steam"` and they pick glyphs accordingly - same mechanism as our built-ins, no separate escape hatch to learn. That's the intended shape of "KAPLAY provides sane defaults, doesn't own the whole problem."

I have pushed a change in line with the above, still happy to revise it though but at least gives us something we can point at and say not this or like that instead how this.

CEREBR4L:
Think this is as far as I would push this PR and happy to even split it out as it feels there is a feature addition conflated a little with a bug fix here but happy to push it through at the size it is as is, if that is acceptable. Assuming the view is we are just improving the controller/gamepad compatibility and functionality. 

When I say as far as I will push it I mean in terms of features and what it addresses, happy to make changes and be wrong but don't want more to creep in as will become hard to review anway.

dragoncoder047 (approved):

src/types.ts
```diff
@@ -196,13 +196,40 @@ export type ChordedKGamepadButton =
  */
 export type KGamepadStick = "left" | "right";
 
+/**
+ * A machine-readable controller family, for picking button-glyph assets.
+ * Not a closed set - custom `GamepadDef`s (via `KAPLAYOpt.gamepads`) can use
+ * any string here, e.g. `"steam"` for a Steam Controller mapping KAPLAY
+ * doesn't ship a built-in for.
```
lajbel: Generation seems obvious here, would you use something more self explanatory

"Available controller families, useful for picking button-glyph assets.

src/types.ts
```diff
@@ -196,13 +196,40 @@ export type ChordedKGamepadButton =
...
+ * @subgroup Gamepad
+ */
+export type GamepadType =
+    | "ps4"
+    | "ps5"
+    | "playstation"
+    | "xbox"
+    | "switch"
+    | (string & {});
+
```
lajbel, line 213: Also I'm not sure it being a non restricted union. Maybe they should be like that when used as parameter, but what if a User want to use it strictly...? I think I would like your opinion here

src/types.ts
```diff
@@ -216,6 +243,13 @@ export type GamepadDef = {
 export type KGamepad = {
     /** The order of the gamepad in the gamepad list. */
     index: number;
+    /**
+     * A human-readable label for the recognized controller, e.g. "DualSense",
+     * or "Standard Gamepad" if unrecognized. For display/debugging only.
+     */
+    controllerName: string;
```
lajbel: Wouldn't make sense use "name" here?

lajbel (changes requested):
LGTM, just a few points

amyspark-ng:
@CEREBR4L

src/types.ts
CEREBR4L: Done, I thought the same while working on it, but I didn't have a strong opinion either way. Now it is done, do think name is better :)

src/types.ts
CEREBR4L: Updated

src/types.ts
CEREBR4L: The reason I opted for it was the flexibility it provides, and it was already a pattern I observed in the key definitions currently line 141 of this file. Mostly I was leaving this open so users could define more custom controllers as we can't maintain an exhaustive list without considerable effort. 

Given its indented function issues with name mismatches and such should be picked up promptly as drives visual assets in some way, my opinion is every code choice is ultimately a trade-off and this feels like it adds more value that it detracts overall for the people building games with the engine. However, happy to change if wanted but I would like to understand the reasoning if possible.

amyspark-ng:
i think this has merge conflicts btw, fix

CEREBR4L:
@amyspark-ng now resolved - wanted to leave close to merge as possible as thought other bits might get merged in interim

amyspark-ng:
still missing review, will have to wait

lajbel (approved):
LGTM

lajbel (approved):

PR #1122, opened 2026-07-07
fix: catch and standardize inside-out raycast normals

imaginarny:
**Fixes various issues with `raycastRect`:**
- when ray is originating inside a shape, no hit is detected
- when casting a parallel ray it's getting caught by bbox when shouldn't (actual shape wouldn't get hit)

| Hit not detected at all | Hit detected, trapped |
|--------|--------|
| rect no hit | rect trapped | 

**General raycasting behavior besides the rect:**
- rays cast from inside a shape are detected but not trapped
  - circle/ellipse detects the hit but ray escapes as the outside normal is used
  - only triangle (polygon) works as should, all will behave like this after the fix

| Hit detected, escapes | Hit detected, trapped | Hit detected, trapped |
|--------|--------|--------|
| ellipse no hit | ellipse fixed | triangle correct | 

Can be compared with existing `examples/raycastShape` and [play.kaplayjs.com/?example=raycastShape](https://play.kaplayjs.com/?example=raycastShape).

## Summary

Fixed various `raycastRect` issues and standardized raycasting across all shapes so rays originating inside a shape are now consistently trapped.

- [x] Changeloged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,9 @@ So your change should look like:
   `obj.pos.x += 1` work again (#1109) - @ErikGXDev
 - Fixed `isKeyDown` and `isButtonDown` getting stuck on game loosing focus
   (#1101) - @Stanko
+- Fixed various `raycastRect` issues and standardized raycasting across all
+  shapes so rays originating inside a shape are now consistently trapped
+  (#1122) - @mflerackers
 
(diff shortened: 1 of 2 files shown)


dragoncoder047:
to be honest, raycasting from inside a shape is undefined behavior, but nice fix so it's at least predictable now

lajbel (approved):

lajbel:
LGTM

PR #1125, opened 2026-07-24
perf: optimize text component

### LINKED ISSUE #1095: feat: make formatted text parsing and rendering more efficient

currently the `text()` component re-parses and re-formats the text on EVERY update, so if you want to render a large amount of formatted text, the best solution is to have the object be `pause`d and then you don't get animated letters (e.g. a formatting function that depends on `time()`). It would be cool if the text could be optimized to only re-*parse* when the text is changed, and then keep the parse data and just recalculate the formatting parameters on each character on update instead.

### Any more information?

I'm making a thing that uses text to render primarily (like 6000+ characters that need to be parsed) and it makes my CPU fans go into turbojet mode if I leave it onscreen for more than like 30 seconds.

imaginarny:
Optimizes the text component performance significantly (by ~30 FPS for large amounts of text).

Fixes #1095 by separating the overall text formatting and transforming (char styles). Text comp now mostly uses transform (`transformFormattedText`), update is called manually when needed and reformat (`formatText`) happens only when required (changes in text, dimensions, renderProp anchor). `onUpdate` now (mostly) just checks for used `renderProps` changes first (can't avoid), so with no changes and text static nothing gets called/updated at all. Unless, there are dynamic text `styles`/`transform` passed, then again, transform will be called only. When there is `stretchInPlace` option set to or evaluates to `false`, it will get reformatted once. Perf is now basically the same as with `txtObj.paused = true` workaround.

Can test all playtests prefixed with **`text`** which should behave the same as before. Added a new **`textPages`** perf playtest with 6300 chars with styles that now maintains 60 FPS, previously it would drop to ~30 FPS and make PC fans go into turbojet mode.

Also, fixed wrong dimensions reported and object not scaled properly along when scale comp used as reported earlier on DC by @hinum. **`textScale`** playtest added for that. Can check the scale issue on [kaplayground](https://play.kaplayjs.com/?code=eJxlUs1u2zAMvuspuKAYZMDImh6GIUELbKcdNuxQ7FQUCCMzklpXMix5ceYZ2EPksOfLk0yS49RBCR9E8jO%2Fjz%2FPWJW45x1sUDzL2jamWMLDdQ7xe8xha41fwkxhVe1nOZRWfsd2CQvosxVjpcXia0zx6BW0aeRcG1eR8HALvm4ogsjDLrjXq%2FRUw5NZ8y38zdHtjQCewe0ddAyCCWucB9%2FGElgU%2FCFFo3lqPT970dbsx%2BYpsu104dUHRVoqD07ZpizAedwzrwgcvhC8B6wJTzkmg5KNlpJq2CkyzAksqYCmeree%2B1q%2F8Cy%2FIOouvGhYamnCaAQZT%2FUsfwNw%2BjeFSX16mym1ofsKhTZyCR8v8%2F2rO5FQWccHIp5NwmiEsjUfNUwyqR%2B%2BmGJD92NXj3Fd8RH3EiY9T%2BNbpZA6hYZZjjjUni%2Fy055iOhEE6C8SN%2FwmGwta87Mq0BNPyDP5cBrheviafaGtrcNkrrpd3151qoc%2F8Hkb9MfQWUxMvcqImPvIOGIS%2Fbzt2fFwOB7%2BwlA0LDlVml4AnC8g4I6Hf%2BssqQqC4wn%2FB8cb1Yo%3D&version=4000.0.0-alpha.27.1) (perf one is too long to share).

## Summary

- Improved `text` component performance by separating text transform and formatting, reducing update calls for both dynamic and (especially) static text
- Fixed objects with a `text` component reporting wrong dimensions when scaled using the `scale` component

- [x] Changeloged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,9 @@ So your change should look like:
   (#1097) - @Stanko
 - **(!)** `new RNG()` and `setRNG()` now use config objects instead of the
   string/custom rng parameter (#1097) - @Stanko
+- Improved `text` component performance by separating text transform and
+  formatting, reducing update calls for both dynamic and (especially) static
+  text (#1125) - @imaginarny
 
 ### Fixed
 
@@ -69,6 +72,8 @@ So your change should look like:
   the initial `GameObjRaw.use()` call (e.g. `obj.use(scale(2))`) - @mflerackers
 - Fixed `isKeyDown` and `isButtonDown` getting stuck on game loosing focus
   (#1101) - @Stanko
+- Fixed objects with a `text` component reporting wrong dimensions when scaled
+  using the `scale` component (#1125) - @imaginarny
 
 ## [4000.0.0-alpha.27.1] - 2026-05-12
 
--- a/src/ecs/components/draw/text.ts
+++ b/src/ecs/components/draw/text.ts
@@ -172,7 +172,7 @@ export function text(t: string, opt: TextCompOpt = {}): TextComp {
     let _width = opt.width ?? 0;
     let _height = 0;
 
-    // obj props that are checked on update and trigger reformat
+    // obj props that are checked on update and trigger transform or reformat
     let _renderProps:
         | RenderProps & Record<"anchor", DrawTextOpt["anchor"]>
         | null = null;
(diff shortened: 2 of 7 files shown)


lajbel (approved):

lajbel:
LGTM

imaginarny:
Ready to merge then @lajbel.

PR #1128, opened 2026-07-30
Fixed test server not being up before its address is requested

Stanko:
When running tests puppeteer couldn't reach the dev server. It took me more time than I'm willing to admit, but I found the error - it was a race condition because `server.address()` is read right after `serve()` and server wasn't broadcasted the "listening" event resulting in `http://localhost:unknown/ai`

The fix is to wait for the "listening" event before proceeding.

- [x] Changeloged

### DIFF
--- a/scripts/test.ts
+++ b/scripts/test.ts
@@ -1,6 +1,7 @@
 // @ts-check
 
 import fs from "fs";
+import { once } from "node:events";
 import path from "path";
 import puppeteer from "puppeteer";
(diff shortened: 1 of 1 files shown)


lajbel:
Hi @Stanko, this feature doesn't need changelog!

Stanko:
Makes sense, removed it.

lajbel (approved):

lajbel:
LGTM

PR #1133, opened 2026-08-03
fix: Restored fixed canvas sizing and added clientToViewport

Stanko:
@imaginarny these are my suggested changes for #1106

- Restored fixed canvas sizing
- Respored aspect ratio fitting when letterbox is passed
- Added `clientToViewport` helper for both touch and mouse pos
- Handled `object-fit: contain` which I use, but it is also used in user agent stylesheets when element is put in fullscreen 

Important thing is that it still doesn't calculate properly in some edge cases like:
- Adding border or padding to the canvas element itself
- Using object-fit `cover` or `scale-down`

But I think it is fine as these are not real scenarios for a game canvas. It should be mentioned in the docs though.

EDIT: Full disclosure - I used LLM for generating the tests and then cleaned them up by hand

## Contributor Checks

- [x] I'm aware of the

## Reviewer Checks

- [ ] Changes include e2e tests
- [ ] Changes include unit tests

### DIFF
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -20,7 +20,7 @@ import type {
 } from "../events/eventMap";
 import { type KEventController, KEventHandler } from "../events/events";
 import { canvasToViewport } from "../gfx/viewport";
-import { map, vec2 } from "../math/math";
+import { vec2 } from "../math/math";
 import { Vec2 } from "../math/Vec2";
 import { _k } from "../shared";
 import { deprecateMsg } from "../utils/log";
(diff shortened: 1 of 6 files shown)


imaginarny:
Logic isn't wrong, but it's again just symptoms patching as with canvasScale previously. Even on each move, which was something I also wanted to avoid, since it could be just set up correctly once in the first place. Basically viewport's work doubled. So can't merge this, but don't fret, I will revert the logic partially to restore that fixed size canvas buffer won't resize, and not only that. Will also include an opt that will enforce that when used with letterbox opt (which makes canvas take 100% size), so no outside setup will be needed. But you won't even need that since you got it pretty custom with your divs setup. So it's a good news, it will work for you as before. More on that in the parent PR. 🙂

Stanko:
Don't worry, I never intended for this to get merge, more to show where I think the painpoints are.

PR #1135, opened 2026-08-06
feat: expose fixURL to the context

### LINKED ISSUE #1084: feat: expose fixURL

plugins that define custom assets or monkeypatch/hook into existing asset types could benefit from being able to access `fixURL()` to get a normalized URL with `window.kaplayjs_assetsAliases` and `globalOpt.urlPrefix` support

### Any more information?

_No response_

amyspark-ng:
# PR Overview

Simply exposes the fixURL function to solve an issue

### List of changes

- Closes #1084 

## Contributor Checks

- [x] I'm aware of the

## Reviewer Checks

- [x] Changes include a changelog


src/core/contextType.ts
```diff
@@ -3412,6 +3412,12 @@ export interface KAPLAYCtx {
      * @subgroup Util
      */
     loadRoot(path?: string): string;
+    /**
+     * Adds the current asset root and urlPrefix to a passed url.
+     *
+     * @param url - The url to add the root at the start of.
+     */
+    fixURL<D>(url: D): D;
```
lajbel: Why using generic here?

lajbel (changes requested):

src/core/contextType.ts
dragoncoder047: i think if it's some string subset, the result is also can be considered to be that string subset

```suggestion
    fixURL<D extends string>(url: D): D;
```

^should be this anyway

amyspark-ng:
oops

src/core/contextType.ts
lajbel: No. That is not true, if we pass url

"bean" -> `fixUrl()` -> "prefix/bean" then D as parameter and D as return is not correct.

lajbel (changes requested):

src/core/contextType.ts
amyspark-ng: Is the generic even needed, can't we just do string

amyspark-ng: ok so, what was this about? can't we just do string?

PR #1136, opened 2026-08-06
feat: add debugButtons

### LINKED ISSUE #920: feat: customize debug keys using buttons api

Have an option like:

```js
kaplay({
  debugButtons: {
      "inspect": {
        keyboard: "f",
        mouse: "left"
      }
  }
});
```

### Any more information?

_No response_

amyspark-ng:
# PR Overview

Adds the debugKeys option to the kaplay opt so the keys used for debug actions can be changed and modified at runtime

If the way it's implemented can be change anyway please tell me i don't know how good this is

### List of changes

- Closes #920 
- Adds debugKeys so keys for debug actions changed, retrieved and modified at runtime
- Now onButtonPress accepts PropertyKey instead of simply string

## Contributor Checks

- [x] I'm aware of the

## Reviewer Checks

- [ ] Breaking changes are marked in the changelog


dragoncoder047 (changes requested):
I would use symbols for the actual button names that go to the buttons system, since "pause" in particular could be a valid button name for normal gameplay, the debug buttons menu should use the well-known string names and then translate internally

amyspark-ng:
okay but like symbols how

dragoncoder047:
would be something like `const DEBUG_SYMBOLS = { pause: Symbol("pause"), inspect: Symbol("inspect"), ...}` and then on line 193 or whatever when you do `opt.debugButtons ?? default`, you instead do `Object.fromEntries(Object.entries(opt.debugButtons ?? default).map(([name, binding]) => [DEBUG_SYMBOLS[name], binding]))`

Also, I would say change this too

```diff
-export type ButtonsDef = Record<string, ButtonBinding>;
+export type ButtonsDef<T = string> = Record<T, ButtonBinding>;
```

and then give the button names in the definition of the global options, so that typescript yells at you if you don't give all the buttons

amyspark-ng:
i think i understand, i've never used javascript's symbols so im gonna have to check

amyspark-ng:
would this allow to changing them at runtime tho? with setButton("inspect", {  keyboard: "f" })

dragoncoder047:
The point of symbols is that they're valid property keys and can be used in place of strings but they're unique, so `Symbol("foo") !== Symbol("foo")`, so if you keep the kaplay debug symbols private then it guarantees no collisions

dragoncoder047:
> would this allow to changing them at runtime tho? with `setButton("inspect", { keyboard: "f" })`

I would make it a separate function, `setDebugButton("inspect", { keyboard: "f" })`, it would have to go through the string->symbol translation table

amyspark-ng:
but if export type ButtonsDef<T = string> = Record<T, ButtonBinding>; doesn't that mean that symbols couldn't be used to index buttonbindings? the symbols would be used when storing debug functions in the buttons object, can't have a different object because then everything button system related would get more complicated

amyspark-ng:
i guess _k.app.state.buttons doesn't HAVE to be specifically a ButtonsDef

amyspark-ng:
now the issue is that in https://github.com/kaplayjs/kaplay/pull/1136/changes#diff-77ec17514ad6254bbf78845b0d2d9b2c332df3f80fc13f186e9479a77f65b9d5R50 i can't `onButtonPress(DEBUG_SYMBOLS["inspect"])` because in regular onButtonPress you can't use symbols as a button name, hmmm

dragoncoder047:
> now the issue is that in https://github.com/kaplayjs/kaplay/pull/1136/changes#diff-77ec17514ad6254bbf78845b0d2d9b2c332df3f80fc13f186e9479a77f65b9d5R50 i can't `onButtonPress(DEBUG_SYMBOLS["inspect"])` because in regular onButtonPress you can't use symbols as a button name, hmmm

you'd probably do it here

https://github.com/kaplayjs/kaplay/blob/8bd9f147d00aa4c36d9609586a560b7a61060cb9/src/app/app.ts#L784

haven't a clue why it's `string | string` anyway, just change one to `symbol` or replace the entire thing with `PropertyKey` which is typescript standard library for `string | symbol`

amyspark-ng:
i don't think it would be as easy since the whole buttons system works with strings only and would have to go through all of it and change it to PropertyKey and even then i don't know if it would still work

amyspark-ng:
Right now i don't know what would be the best way to implement this, since as to have debug functions sepparated from other user actions, i mean

```js
kaplay({
    debugButtons: { pause: { keyboard: "f8" } }
    buttons: { pause: { keyboard: "p" } }
})
```

right now im trying to avoid this, since debugButtons are currently being added to the `_k.app.state.buttons`, but this would allow users to modify existing debugButtons as shown in the example above

debug buttons have to be added to the `_k.app.state.buttons` object because it's the one events such as `onButtonPress` check to make the functions trigger

This is why @dragoncoder047 suggested using symbols internally to differentiate debug keys from regular buttons so
```js
_k.app.state.buttons = {
    Symbol("pause"): { keyboard: "f8" },
    pause: { keyboard: "p" }
}
```

The issue is that on https://github.com/kaplayjs/kaplay/blob/9ec9f57faa95c1a2febdaa3fd3a5e361c363b6fd/src/app/appEvents.ts i would need to call onButtonPress() and have it check for the Symbol() stored debug buttons, which is not possible since Symbols can't be used as a name in the onButtonPress function

This would be fixed by allowing it to use Symbols AND strings (PropertyKey) on the onButtonPress, but this is an issue since all of the button systems is made in mind as having strings as button names as all of the code in the class of  `ButtonProcessor`
https://github.com/kaplayjs/kaplay/blob/9ec9f57faa95c1a2febdaa3fd3a5e361c363b6fd/src/app/inputBindings.ts#L149

so... thoughts? @dragoncoder047
do you think it's worth it to look through all of it and see if it's possible to change all of the button system to see if it works or should we go through other alternative

dragoncoder047:
It should work if you change all to PropertyKey, I just had used string there since that's what buttons were at the time.

amyspark-ng:
a work of genius

amyspark-ng:
now this makes it so onButtonPress accepts PropertyKey instead of only string, i don't know how breaking this change is and i don't know if the PR is completely tested out, please reviewers check

src/app/app.ts
```diff
@@ -185,13 +183,43 @@ export type FixedSpeedOption = keyof typeof fixedSpeeds;
...
+        }),
+    );
+
     return {
         canvas: opt.canvas,
-        buttons: buttons,
+        buttons: { ...buttons, ...debugButtons } as Record<
+            symbol | string,
+            ButtonBinding
+        >,
```
dragoncoder047, line 220: is the 'as' necessary here now that all is PropertyKey?

amyspark-ng: removing it seems to cause these errors, i wouldn't

image

amyspark-ng:
could debug.setButton() and debug.getButton() be more intuitive?

amyspark-ng:
i think it would be more a refactor so i messed up with the commit type oops

amyspark-ng:
im just now seeing this branch

https://github.com/kaplayjs/kaplay/compare/master...customizable-debug

i don't know what the update fixedUpdate and draw things are all about but yeah

src/constants/general.ts
```diff
@@ -52,3 +52,13 @@ export const EVENT_CANCEL_SYMBOL = Symbol.for("kaplay.cancel");
...
+
+// Using symbols so if this is mixed with the app.state.buttons, you can't do setButton("inspect") and modify it
+export const DEBUG_SYMBOLS = {
+    "inspect": Symbol("inspect"),
+    "clearlogs": Symbol("clearlogs"),
+    "pause": Symbol("pause"),
+    "slowdown": Symbol("slowdown"),
+    "speedup": Symbol("speedup"),
+    "stepframe": Symbol("stepframe"),
+};
```
lajbel, line 65: Great

src/core/contextType.ts
```diff
@@ -2745,10 +2745,10 @@ export interface KAPLAYCtx {
      * @subgroup Buttons API
      */
     onButtonDown(
-        btn: string | string[],
-        action: (btn: string) => void,
+        btn: PropertyKey | PropertyKey[],
+        action: (btn: PropertyKey) => void,
     ): KEventController;
-    onButtonDown(action: (btn: string) => void): KEventController;
+    onButtonDown(action: (btn: PropertyKey) => void): KEventController;
```
lajbel, line 2751: Would you explain modifications of this? Have you checked `src/core/taf.ts`?

lajbel (changes requested):

src/constants/general.ts
amyspark-ng: is this constant okay here? should i move it somewhere else maybe on the debug file? why the need to request a change here....

src/core/contextType.ts
amyspark-ng: i had forgotten about taf.ts, but then why don't onButtonPresses work with taf.ts code and they all say string rather than ButtonName? maybe this typing work is beyond me but i have to look more into it then

the reason of the change was so Symbols() could be used as a parameter in btn, since Symbol("inspect") is used in https://github.com/kaplayjs/kaplay/pull/1136/changes#diff-77ec17514ad6254bbf78845b0d2d9b2c332df3f80fc13f186e9479a77f65b9d5R51

lajbel:
Please fix breaking changes

amyspark-ng:
yes mam

amyspark-ng:
still have to make it work with taf.ts (?) and change PropertyKey type param on onButtonPress to inline's String | Symbol

PR #1140, opened 2026-08-09
feat: AudioPlayOpt.reverse

### LINKED ISSUE #465: feat: Play sounds in reverse
add an opt so you can play a sound in reverse

amyspark-ng:
# PR Overview

Adds the option to reverse an audio when playing it

### List of changes

- Closes #465 
- Adds AudioPlayOpt.reverse so a sound can be reversed when playing it

## Contributor Checks

- [x] I'm aware of the

## Reviewer Checks


src/audio/play.ts
```diff
@@ -183,7 +189,27 @@ export function play(
...
+                data.buf.sampleRate,
+            );
+
+            // Loop through each channel and reverse the data
+            for (let i = 0; i < data.buf.numberOfChannels; i++) {
+                const originalData = data.buf.getChannelData(i);
+                const reversedData = reversedBuffer.getChannelData(i);
+
+                // Copy and reverse
+                reversedData.set(originalData.slice().reverse());
```
dragoncoder047:
```suggestion
                reversedData.set(originalData.toReversed());
```

lajbel: Would you apply this? @amyspark-ng

examples/audio.js
```diff
@@ -45,6 +53,7 @@ onTouchStart(() => {
 const music = play("OtherworldlyFoe", {
     loop: true,
     paused: true,
+    reverse: true,
```
lajbel: This seems like a test. reverse doesn't work with Music.

- If it is a streaming limitation, I prefer separating `play()` and a new `playMusic()`
- If we can reverse the audio from `loadMusic()`, anyway I still prefer separating `play()` and a new `playMusic()` for music-only reproduction, with their own options object so we don't confuse users

I don't say we should do it in this PR but you can create a new issue for these changes post merge #1140.

lajbel: So please, remove this line

src/audio/play.ts
```diff
@@ -42,6 +42,12 @@ export interface AudioPlayOpt {
      * ```
      */
     detune?: number;
+    /**
+     * Wheter to reverse the sound, so it plays from finish to start, this doesn't affect time().
```
lajbel, line 46: How possible and what ETA we would have to implement a `trueTime()` or a way to know the time of the sound is being played (for example, if we are reversing and we are in the 80% of the song, time() would be 2 sec (in a 10sec song) since it is reversed

Seems confusing... but I guess you can get me.

lajbel (changes requested):
LGTM, just a few points

examples/audio.js
amyspark-ng: Yeah, i was just testing

It is a streaming limitation, rooted in the way how music is <audio> and regular play() isn't, this is a big semi-breaking change i think? i've been thinking what to do with that for a while kinda

Should we migrate all sounds to <audio> since it seems to be faster than regular sounds? or is it not suited for sounds, what should we do, i will open an issue on this topic i think

src/audio/play.ts
amyspark-ng: i mean, time() right now works as trueTime() as you would say i think, since even though the song is reversed is still 2 seconds into playing, so nothing changes, i think something like trueTime() would apply more when changing the speed, of the sound, but if the speed changes you would want time() to change too i think, so i don't see the need

examples/audio.js
amyspark-ng: https://github.com/kaplayjs/kaplay/discussions/1141 was opened

lajbel (approved):

PR #1156, opened 2026-08-25
test: from pupeteer/vitest to playwright + new testing organization

lajbel:
# PR Overview

Ohhiiiiii :>

## List of changes

- Adds a new testing framework, playwright
- Workflows are adapted
- Vitest and Puppeteer dependencies are removed
- Now tests are ran by Playwright using `npm run test`
- Now there's a folder for `integration/` tests (tests that loads kaplay()) and `unit/` tests (tests that uses the raw functions from the engine)
- Some tests were moved to `.archive` folder in case to need to be redone or smt like that, like type tests, that will be in another PR
- Modificates PR template reflecting these changes
- Workflows still works, playtests are still available in npm run dev and it keeps reading `tests/playtest` and `examples/`

### Future changes?

- TypeScript tests
- Some playtests automated through integration tests
- Revamp playtest views, browsing, catalog and workflows related to the playtests process

## Contributor Checks

- [x] I'm aware of the [contributing guidelines](https://github.com/kaplayjs/kaplay/blob/master/CONTRIBUTING.md).

## Reviewer Checks

- [ ] Breaking changes are marked in the changelog
- [ ] Changes include e2e tests
- [ ] Changes include unit tests

## Changelog

Not needed since it is a testing PR.


imaginarny (commented):
Seems ok, but have the same feedback/question as the last time. Why omit the existing grouping descriptions? It adds more context and can be done the same way as previously by wrapping, just the fn changed from `describe(...)` to `test.describe(...)`. Haven't compared them all, so usefulness might vary, but might be worth checking.

E.g. `tests/unit/math/color.spec.ts`:
```diff
import { expect, test } from "@playwright/test";
import { Color, rgb } from "../../../src/math/color";

+ test.describe("Color creation using class", () => {
    test("new Color(44, 44, 44) should return Color(44, 44, 44)", () => {
        const color = new Color(44, 44, 44);
        expect(color).toEqual({ r: 44, g: 44, b: 44 });
    });

    test("Color().fromArray([22, 43, 79]) should return Color(22, 43, 79)", () => {
        const color = Color.fromArray([22, 43, 79]);
        expect(color).toEqual({ r: 22, g: 43, b: 79 });
    });

    test("Color().fromHex(0x123456) should return Color(0x12, 0x34, 0x56)", () => {
        const color = Color.fromHex(0x123456);
        expect(color).toEqual({ r: 0x12, g: 0x34, b: 0x56 });
    });

    test("Color().fromHex('#234567') should return Color(0x23, 0x45, 0x67)", () => {
        const color = Color.fromHex("#234567");
        expect(color).toEqual({ r: 0x23, g: 0x45, b: 0x67 });
    });
+ });

+ test.describe("Color creation using rbg() utility", () => {
    test("rgb(255, 255, 255) should return Color(255, 255, 255)", () => {
        const color = rgb(255, 255, 255);
        expect(color).toEqual({ r: 255, g: 255, b: 255 });
    });

    test("rgb(11, 25, 6) should return Color(11, 25, 6)", () => {
        const color = rgb(11, 25, 6);
        expect(color).toEqual({ r: 11, g: 25, b: 6 });
    });
+ });

```

Also, the frst time running tests it will flood with errors. `npx playwright install` has to be run first. But I wouldn't include it as a post install script as it's heavy and not needed when installed as a submodule and so on (if so, only as pretest with chromium executable path check). But I don't know if it's worth the hassle, it is mentioned after each error in console and we currently don't instruct contributors to run tests manually anyway. Should we actually (depending on changes)? Mentioned just in case.

imaginarny (approved):

PR #1121, opened 2026-07-06
fix: set area scale properly and allow vec2 prop changes on scale and offset

amyspark-ng:
## Please describe what issue(s) this PR fixes.
- fixes #1120
- allows `Vec2` prop mutations on `scale` and `offset` again

## Summary

- [x] Changeloged

### Fixed
- Fixed `AreaComp.area.scale` not being set properly and restored support for
  `Vec2` property mutations on `scale` and `offset` (#1121) - @amyspark-ng,
  @imaginarny

PR #1161, opened 2026-09-03
chore: add changelog for #1121

amyspark-ng:
# PR Overview

slight mishap in changelogs meant #1121 was merged but its changelog wasn't added, fixed here


dragoncoder047:
this is why I never supported putting the changelog in the PR overview and not in the actual changelog file!!

imaginarny:
This is why nothing will ever work since nobody reads or at least not with comprehension. I even messaged you I updated the changelog. Lajbel merging without it was mistake too, the very same cause. Existing PRs were never meant to remove changelog in the first place and.. well, just no point to continue.

Changelog is already there, up with all related vec2 proxy issues fixed.

amyspark-ng:
oop

PR #1163, opened 2026-09-11
Add a smooth/anti-aliasing rendering option

scambier:
# PR Overview

This PR adds a third option to `KAPLAYOpt.crisp`: `"smooth"`. This will apply a shader that adds anti-aliasing to pixel edges, while maintaining their crispiness. This is especially noticeable on labels, thin lines, and more generally "big pixel" sprites that are rendered unevenly when the scaling is not an integer.

Disclaimer: this was mostly written by AI, out of a personal need. I understand what's going on but rendering and shaders aren't my thing at all, so I absolutely won't take it personally if it's rejected because of it :)

## List of changes

- New setting value for `KAPLAYOpt.crisp`
- New anti-aliasing shader + related demo page and basic integration tests
- This new render option needs to be integrated relatively deeply, but the changes are (I think) quite straightforward.

Comparison screenshots:

crisp = "smooth"
image

crisp = false
image

## Contributor Checks

- [x] I'm aware of the

## Reviewer Checks

**Don't modify this section.** It is for reviewers to check when reviewing the PR.

- [ ] Changes include a changelog.
- [ ] Breaking changes are marked in the changelog.
- [ ] Changes include `/tests/integration` tests. 
- [ ] Changes include `/tests/unit` tests.
- [ ] Additional reviewers are assigned to the PR.

### DIFF
--- a/src/app/appEvents.ts
+++ b/src/app/appEvents.ts
@@ -34,6 +34,9 @@ export function initAppEvents() {
                 _k.gfx.ggl,
                 _k.gfx.ggl.gl.drawingBufferWidth,
                 _k.gfx.ggl.gl.drawingBufferHeight,
+                _k.globalOpt.crisp === "smooth"
+                    ? { filter: "linear" }
+                    : undefined,
             );
 
             _k.gfx.width = _k.gfx.ggl.gl.drawingBufferWidth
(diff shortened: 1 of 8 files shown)


lajbel:
Hi @scambier, thank you so much for making your first PR!

It will be discussed by the internal development team.

imaginarny:
The implementation itself isn't correct for general use, so can't merge as you already thought. But shaders are one of the ways used to deal with stuff like this, so good for you and your use case if you are fine with it and good reference to have. We also looked into rendering issues recently and proper general solution is more complicated, at least the one for general use without shortcomings.

scambier:
Great, I'll make sure to check once you have a proper solution :) Have a nice day

PR #1164, opened 2026-09-11
fix: fixed sprite component returning stale src

Stanko:
# PR Overview

This was the problem this PR fixes:
```ts
const obj = k.add([k.sprite("bean")]);
obj.sprite = "apple";
console.log(obj.sprite); // --> prints "bean"
```

Sprite component would always return `.sprite` property that was set on init. Made sure that is properly updated on both init and later on and handled potential race conditions. Added basic integration tests for it.

## List of changes

- Fixed sprite component returning stale value after changing the sprite value dynamically

## Contributor Checks

- [x] I'm aware of the

## Reviewer Checks

**Don't modify this section.** It is for reviewers to check when reviewing the PR.

- [ ] Changes include a changelog.
- [ ] Breaking changes are marked in the changelog.
- [ ] Changes include `/tests/integration` tests. 
- [ ] Changes include `/tests/unit` tests.
- [ ] Additional reviewers are assigned to the PR.


amyspark-ng:
wait, really? can't believe this went uncaught for so much time then, insane

Stanko:
To be fair, swapping sprites on an object is not something you often do :)

Stanko:
Added changelog. I would added it earlier, but I was confused if we should just write it in the PR or update the markdown file.

amyspark-ng:
changelog in file is good, thank you

